PR #18, opened 2024-04-23
Improve unit tests

jpraet:
* cleanup tests and use AssertJ
* some minor code changes discovered during testing: 
  * InputValidationIssues.requiredInput should not have a value param as it is always null
  * Establishment number -> Establishment unit number
  * preceed -> precede
  * fix InputValidationIssue equals and hashcode
  * add .toString() methods

### DIFF
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
@@ -108,10 +108,10 @@ public class InputValidationIssues {
                 .in(in, name, value);
     }
 
-    public static InputValidationIssue requiredInput(InEnum in, String name, Object value) {
+    public static InputValidationIssue requiredInput(InEnum in, String name) {
         return new InputValidationIssue(ISSUE_TYPE_REQUIRED_INPUT, "Input is required in this context")
                 .detail(String.format("Input %s is required in this context", name))
-                .in(in, name, value);
+                .in(in, name, null);
     }
 
     public static InputValidationIssue replacedSsin(InEnum in, String name, String ssin, String newSsin) {
@@ -167,7 +167,7 @@ public class InputValidationIssues {
     public static InputValidationIssue invalidEstablishmentUnitNumber(InEnum in, String name,
             String establishmentUnitNumber) {
         return invalidStructure(in, name, establishmentUnitNumber,
-                String.format("Establishment number %s is invalid", establishmentUnitNumber));
+                String.format("Establishment unit number %s is invalid", establishmentUnitNumber));
     }
 
     public static InputValidationIssue exactlyOneOfExpected(List<Input<?>> inputs) {
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
@@ -67,6 +67,12 @@ public class InputValidationIssues {
                         .in(in, name, value);
     }
 
+    public static InputValidationIssue unknownInput(InEnum in, String name, Object value) {
+        return new InputValidationIssue(ISSUE_TYPE_UNKNOWN_INPUT, "Unknown input")
+                .detail(String.format("Input %s is unknown", name))
+                .in(in, name, value);
+    }
+
     public static InputValidationIssue invalidStructure(InEnum in, String name, Object value, String detail) {
         return new InputValidationIssue(ISSUE_TYPE_INVALID_STRUCTURE,
                 "Input value has invalid structure")
@@ -139,14 +145,14 @@ public class InputValidationIssues {
 
     public static InputValidationIssue invalidPeriod(InEnum in, String name, Object period) {
         return new InputValidationIssue(ISSUE_TYPE_INVALID_PERIOD, "Period is invalid")
-                .detail("endDate should not preceed startDate")
+                .detail("endDate should not precede startDate")
                 .in(in, name, period);
     }
 
     public static <T extends Temporal & Comparable<? super T>> InputValidationIssue invalidPeriod(
             Input<T> start, Input<T> end) {
         return new InputValidationIssue(ISSUE_TYPE_INVALID_PERIOD, "Period is invalid")
-                .detail(String.format("%s should not preceed %s", end.getName(), start.getName()))
+                .detail(String.format("%s should not precede %s", end.getName(), start.getName()))
                 .inputs(Arrays.asList(start, end));
     }
 
--- a/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssuesTest.java
+++ b/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssuesTest.java
@@ -25,6 +25,18 @@ class InputValidationIssuesTest {
     }
 
     @Test
+    void unknownInput() {
+        InputValidationIssue issue =
+                InputValidationIssues.unknownInput(InEnum.BODY, "oops", "value");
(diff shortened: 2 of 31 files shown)


belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/RefDataCollectionValidatorTest.java
```diff
@@ -2,66 +2,38 @@
...
+                .validate()).isEmpty();
     }
 
     @Test
-    void validateNullInput() {
-        assertThrows(NullPointerException.class,
-                () -> new RefDataCollectionValidator<>(null, Collections.singletonList("ok")),
-                "Should have thrown exception");
+    void nokNullAllowedRefData() {
+        assertThatNullPointerException().isThrownBy(
```
clone1612, line 23: Maybe check exception message here as well?

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/RefDataCollectionValidatorTest.java
```diff
@@ -2,66 +2,38 @@
...
     }
 
     @Test
-    void validateNullIn() {
-        assertThrows(NullPointerException.class,
-                () -> new RefDataCollectionValidator<>(new Input<>(null, "refData", "ok"),
-                        Collections.singletonList("ok")),
-                "Should have thrown exception");
+    void nokEmptyAllowedRefData() {
+        assertThatIllegalStateException().isThrownBy(
```
clone1612, line 29: Maybe check exception message here as well?

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/RefDataPredicateValidatorTest.java
```diff
@@ -2,56 +2,29 @@
...
-                "Should have thrown exception");
-    }
-
-    @Test
-    void validateNullAllowedRefData() {
-        assertThrows(NullPointerException.class,
-                () -> new RefDataPredicateValidator<>(new Input<>(QUERY, "refData", "ok"), null),
-                "Should have thrown exception");
+    void nokNullAllowedRefData() {
+        assertThatNullPointerException().isThrownBy(
```
clone1612, line 20: Maybe check exception message here as well?

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/RequestValidatorTest.java
```diff
@@ -19,302 +17,432 @@
...
+    }
+
+    @Test
+    void zeroOrAllOfValidAll() {
+        Input<?>[] inputs = { Input.query("a", "ok"), Input.query("b", "ok") };
+        assertValid(new RequestValidator().zeroOrAllOf(inputs));
+    }
+
+    @Test
+    void zeroOrallOfInvalid() {
```
clone1612, line 201: Nitpick: all should be capitalized

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/RequestValidatorTest.java
```diff
@@ -19,302 +17,432 @@
 import io.github.belgif.rest.problem.api.InputValidationIssues;
 
 class RequestValidatorTest {
```
clone1612, line 19: There doesn't seem to be a test which chains multiple validations and verifies they are all checked (i.e. validation does not fail fast)?

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/TimeReflectionUtilTest.java
```diff
@@ -49,20 +49,27 @@ void nokWithInvalidType() {
                 "Should have thrown exception");
     }
```
clone1612, line 51: This test class still uses JUnit assertions for validating an exception was thrown

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/validation/ZeroOrAllOfValidatorTest.java
```diff
@@ -1,59 +1,36 @@
...
 import java.util.List;
 
 import org.junit.jupiter.api.Test;
 
 import io.github.belgif.rest.problem.api.Input;
-import io.github.belgif.rest.problem.api.InputValidationIssue;
 import io.github.belgif.rest.problem.api.InputValidationIssues;
 
 class ZeroOrAllOfValidatorTest {
     private ZeroOrAllOfValidator tested;
```
clone1612, line 14: This field is no longer used

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
```diff
@@ -108,10 +108,10 @@ public static InputValidationIssue rejectedInput(InEnum in, String name, Object
                 .in(in, name, value);
     }
```
clone1612, line 110: The class also contains `ISSUE_TYPE_UNKNOWN_INPUT` but it isn't used in a method nor does it have a test. Is that normal?

clone1612 (changes requested):
A very nice increase in test quality/coverage, I've added some minor remarks.

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssuesTest.java
```diff
@@ -0,0 +1,303 @@
...
+
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+class InputValidationIssuesTest {
```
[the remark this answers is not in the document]
jpraet: You mean something like this?

```java
        assertThat(InputValidationIssues.schemaViolation(InEnum.BODY, "test", "value", "detail"))
                .isEqualTo(new InputValidationIssue(ISSUE_TYPE_SCHEMA_VIOLATION,
                        "Input value is invalid with respect to the schema")
                                .detail("detail")
                                .in(InEnum.BODY, "test", "value"));
```

My concern with that is that the test code would then basically do the same as the production code: construct the expected InputValidationIssue. It is likely that such test would be created starting by copying the production code into the test, and potentially just duplicating a bug in the test?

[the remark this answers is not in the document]
clone1612: That is indeed a valid concern, what about using the no-args constructor with the setters (not the builder methods)? That way we also test the different styles of constructing the issue. Or would that make the tests too verbose?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
jpraet: It's a standardized issue type at Belgif (https://www.belgif.be/specification/rest/api-guide/#rule-req-valid) which we didn't use at CBSS so far.

I added the corresponding InputValidationIssues.unknownInput() method and a unit test.

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssuesTest.java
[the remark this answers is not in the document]
jpraet: I have added assertions to verify the empty properties:

```java
 assertThat(issue).extracting("href", "instance", "status", "inputs", "additionalProperties")
        .allMatch(this::isEmpty);
```

clone1612 (approved):

PR #22, opened 2024-04-24
Extract RequestValidator to belgif-rest-problem-validator

### LINKED ISSUE #16: Extract validators
The library contains a list of input validators (e.g. validate period end after start, SSIN checksum, ...).

Smals uses another library for validation, based on Bean Validation annotations: https://www.ict-reuse.be/nl/service/smalsutils-validation . These annotations are added to the generated code by adding "x-[...]-extra-annotation" properties in the OpenAPI document.

I suggest to:
* define a standard way how input validation errors can be propagated in a way the rest-problem-java library can handle them
* extract validators out of the rest-problem-java library
* (optional) refactor/improve a common validation library

jpraet:
Fixes #16.


belgif-rest-problem-validator/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
```
clone1612, line 36: Also used in `belgif-rest-problem-java-ee`, could we move it to the root pom?

clone1612 (approved):

belgif-rest-problem-validator/pom.xml
jpraet: That seems to be causing issues for the dependencies in the spring modules.

Apparently dependencyManagement in parent pom.xml takes precedence over the imported spring-boot-dependencies BOM.

PR #25, opened 2024-04-26
Move additionalProperties from DefaultProblem to Problem

### LINKED ISSUE #1: Move additionalProperties from DefaultProblem to Problem
An additionalProperties with JsonAnyGetter/JsonAnySetter is currently defined on DefaultProblem (the fallback class for unmapped problem types).

It would be better if these additionalProperties would be pulled up into the Problem base class, so we can handle properties that have not been defined in the schema for any problem.

But this requires upgrading to jackson-databind 2.16.2 because of https://github.com/FasterXML/jackson-databind/issues/4316, and I don't think we want to tie the rest-problem-java library to a specific jackson version.

jpraet:
Fixes #1.

Also adds integration tests for jackson (de)serialization:
* with the version 2.12.7 as used on JBoss EAP 7.4
* with the latest version 2.17.0


belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/AbstractJacksonSerializationTest.java
```diff
@@ -1,80 +1,93 @@
...
+        print("jackson version: " + PackageVersion.VERSION);
+    }
+
     @BeforeEach
     void setUp() {
         mapper = new ObjectMapper();
         mapper.enable(SerializationFeature.INDENT_OUTPUT);
         TestProblemTypeRegistry registry = new TestProblemTypeRegistry();
         registry.registerProblemType(BadRequestProblem.class);
+        registry.registerProblemType(CustomProblem.class);
```
clone1612, line 40: varargs method so this can be a single call

clone1612 (changes requested):

clone1612 (approved):

PR #27, opened 2024-04-26
Remove code duplication by introducing AbstractRestProblemIT class

jpraet:
Depends on #25.


belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/AbstractRestProblemIT.java
```diff
@@ -0,0 +1,114 @@
...
+        getSpec().when().get("/retryAfter").then().assertThat()
+                .statusCode(503)
+                .header("Retry-After", "10000")
+                .body("type", equalTo("urn:problem-type:belgif:serviceUnavailable"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("getClients")
+    void badRequestFromBackend(String client) {
+        getSpec().when().get("/badRequestFromBackend?client=" + client).then().assertThat()
```
clone1612, line 65: I find it more readable when the `queryParam(...)` method is used. Now I always have to parse the entire string, look where ?/= are placed, ...

```
        getSpec().when()
                .queryParam("client", client)
                .get("/badRequestFromBackend")
```

What do you think?

clone1612 (changes requested):

clone1612 (approved):

PR #29, opened 2024-04-29
Add documentation and integration test for code generator configuration

jpraet:
Depends on #27.


belgif-rest-problem-it/belgif-rest-problem-codegen-it/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+        <executions>
+          <execution>
+            <id>unpack-belgif-openapi</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>unpack-dependencies</goal>
+            </goals>
+            <configuration>
+              <includeTypes>zip</includeTypes>
+              <outputDirectory>${project.basedir}/src/main/resources</outputDirectory>
```
clone1612, line 40: Can't we extract the files in the target directory? That way we also don't need to worry about the .gitignore

belgif-rest-problem-it/belgif-rest-problem-codegen-it/src/test/java/io/github/belgif/rest/problem/codegen/OpenApiGeneratorIT.java
```diff
@@ -0,0 +1,19 @@
+package io.github.belgif.rest.problem.codegen;
+
+import static org.assertj.core.api.Assertions.*;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class OpenApiGeneratorIT {
```
clone1612, line 8: Should we maybe add a test that validates the config actually generated the 2 interfaces + 1 class for our definition? Exists or attempt to use them, both could work.

This would make sure that our configured mapping doesn't interfere with the generation of other classes.

clone1612 (changes requested):

belgif-rest-problem-it/belgif-rest-problem-codegen-it/pom.xml
jpraet: The openapi.yaml in src/main/resources refers to these extracted files.

clone1612 (approved):

PR #32, opened 2024-05-03
Add bean validation support to belgif-rest-problem-spring

### LINKED ISSUE #12: Add Bean Validation support to belgif-rest-problem-spring
At CBSS, we don't (currently) use Spring Boot for implementing REST APIs. So the belgif-rest-problem-spring library is not completely on par with the belgif-rest-problem-java-ee library.

One of the lacking features is [Bean Validation](https://docs.spring.io/spring-framework/reference/core/validation/beanvalidation.html) support.

More specifically, an ExceptionMapper for converting whatever exception the Spring Framework throws in case of a Bean Validation constraint violation, to a BadRequestProblem, similar to what is done in belgif-rest-problem-java-ee with
[ConstraintViolationExceptionMapper](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/jaxrs/ConstraintViolationExceptionMapper.java) and [ConstraintViolationUtil](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/internal/ConstraintViolationUtil.java).

smals-mavh:
Fixes #12.

Implemented basic bean validation.
Unfortunately the generic beanValidation IT doesn't work for Spring because Spring throws different exceptions for different violations in different phases. For this reason I think it's usefull to test the Spring validation itself extensively in the Integration Tests.
I've put some comments in the Spring IT to provide a bit of explanation.

@jpraet could you please review and provide feedback?

@pvdbosch pointed out that we can use https://github.com/zalando/problem-spring-web/tree/main/problem-spring-web this library as an example.
Then we should also implement handlers for:
(routing)
- MissingServletRequestPartException
- NoHandlerFoundException
- ServletRequestBindingException
(validation)
- InvalidRequestException

In the future we could also implement the default Spring Security exceptions.


smals-mavh:
@jpraet, I'll do some more improvements, and let you know when I'm done.

smals-mavh:
@jpraet , Could you please review?

jpraet:
I made the BeanValidationExceptionsHandling conditional on jakarta.validation.ConstraintViolationException.
So we can keep supporting Spring Boot 2.x. There is now a separate IT for Spring Boot 2.x and 3.x.

jpraet:
By using the same approach as the rest-problem-java-ee library (code against javax.validation, and use org.eclipse.transformer:transformer-maven-plugin to generate a jakarta version of the artifact), we can support bean validation on Spring Boot 2.x as well.

On Spring Boot 3.x you would need to use `<classifier>jakarta</classifier>` then.

I will make a stacked PR on top of this PR to make those changes and we can decide whether we want to go with that or not.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/internal/DetermineSourceUtil.java
```diff
@@ -0,0 +1,105 @@
...
+        return InEnum.QUERY;
+    }
+
+    public static InEnum determineSource(ConstraintViolation<?> violation,
+            LinkedList<Node> propertyPath, MethodNode methodNode) {
+        if (propertyPath.getLast().getKind() == ElementKind.PARAMETER) {
+            if (methodNode != null) {
+                ParameterNode param = propertyPath.getLast().as(ParameterNode.class);
+                try {
+                    Method method = violation.getRootBeanClass().getMethod(methodNode.getName(),
```
jpraet, line 80: In the ConstraintViolationUtil for Java EE, we also look for annotations on superclass or interface.
Not 100% sure if it's needed.

smals-mavh: Good catch, working on it.

smals-mavh: Implemented check of superclasses and interfaces as well.
Added testcase to verify if everything also works if there is an overwrite in the childclass.

jpraet: It didn't traverse the full hierarchy for parent classes / interfaces. Neither did the existing JAX-RS implementation.

I extracted that common logic to an internal AnnotationUtil and fixed it.

jpraet (approved):

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/internal/AnnotationUtilTest.java
```diff
@@ -0,0 +1,127 @@
...
+    void byName() {
+        Optional<? extends Annotation> annotation =
+                AnnotationUtil.findParamAnnotation(getMethod("ownMethod"), "i", Foo.class);
+        assertThat(annotation).isPresent();
+    }
+
+    @Test
+    void byNameFallback() {
+        Optional<? extends Annotation> annotation =
+                AnnotationUtil.findParamAnnotation(getMethod("ownMethod"), "fallback", Foo.class);
```
clone1612, line 46: This fallback behavior isn't described in the javadoc, reading it I was under the impression that an empty `Optional` would be returned but that isn't the case.

clone1612 (changes requested):

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/internal/AnnotationUtilTest.java
jpraet: This fallback was needed to get the nestedQueryParamsViolation and doubleNestedQueryParamsViolation tests working.
In those tests, it's not really the method parameter name that gets passed, but the field name
(from io.github.belgif.rest.problem.spring.BeanValidationExceptionsHandler#handleBindException). Maybe it can be handled differently.

For now, describing this fallback in the javadoc.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/BeanValidationExceptionsHandler.java
```diff
@@ -0,0 +1,83 @@
...
+@Order(1)
+@ConditionalOnClass(ConstraintViolationException.class)
+public class BeanValidationExceptionsHandler {
+
+    @ExceptionHandler(ConstraintViolationException.class)
+    public ResponseEntity<Problem> handleConstraintViolationException(ConstraintViolationException exception) {
+        return ProblemMediaType.INSTANCE.toResponse(
+                new BadRequestProblem(exception.getConstraintViolations().stream()
+                        .map(BeanValidationExceptionUtil::convertToInputValidationIssue)
+                        .sorted(Comparator.comparing(InputValidationIssue::getName))
```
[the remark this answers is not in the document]
jpraet: When implementing Comparable, a compareTo() method that only checks the name would be inconsistent with equals.
Adding a reusable static BY_NAME comparator instead.

belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/AbstractRestProblemSpringBootIT.java
```diff
@@ -0,0 +1,190 @@
...
+    @Test
+    void headerParamTypeMismatch() {
+        getSpec().when().header(new Header("id", "myId")).get("/constraintViolationHeader").then().assertThat()
+                .statusCode(400)
+                .body("type", equalTo("urn:problem-type:belgif:badRequest"))
+                .body("issues[0].in", equalTo("header"))
+                .body("issues[0].detail", equalTo("id should be of type int"))
+                .body("issues[0].value", equalTo("myId"));
+    }
+
```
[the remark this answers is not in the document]
jpraet: Test added, and had to implement a new exception mapper for MissingRequestHeaderException.

clone1612 (approved):

jpraet:
@smals-mavh @pvdbosch OK to merge?

PR #36, opened 2024-05-06
Support bean validation on Spring Boot 2.x

jpraet:


jpraet:
> @jpraet seems indeed like a good solution to support both Spring boot 2 and 3. For development this is a bit annoying because IntelliJ doesn't run the tests for SB3 (it only loads the javax version in the session). But with maven it's no problem to run the tests of course. 

Yes that's annoying. I thought maybe "Unload Maven Module" would do the trick (that it would then resolve the artifact from the local maven repo), but it doesn't seem to be the case.

> I think the merge with #32 will be a bit nasty. Let me know if you'll do it, or if I can do it. (You might want to review the last changes I've made to #32)

The merge is done.

PR #56, opened 2024-05-30
Fix JAX-RS proxy for methods from Configurable interface

### LINKED ISSUE #55: JAX-RS ProblemSupport does not correctly proxy Configurable
* javax.ws.rs.client.Client extends `Configurable<Client>`
* javax.ws.rs.client.WebTarget extends `Configurable<WebTarget>`

Objects returns from calls to the methods from Configurable that return the generic type (Client / WebTarget) are not proxied.

jpraet:
Fix #55


clone1612 (approved):
I've validated the fix in our affected project, looks good!

PR #60, opened 2024-06-03
Add requireIfPresent validator

### LINKED ISSUE #57: Support more complex input relation checks in RequestValidator
Consider an API with query parameters sector & institution that have the following relation:
* Both can be absent
* If institution is provided sector should be as well
* Sector can be provided without an institution

The current methods from `RequestValidator` cannot express this, the closest we have is `when(boolean condition, Consumer<RequestValidator> requestValidatorConsumer)` but that doesn't result in a response which expresses the relation. 

I believe adding a method for this scenario would be useful, although I don't have an immediate idea for its name. Maybe `requireDependency` ?

jpraet:
Fix #57.


clone1612 (approved):

clone1612:
Maybe also add the new validator to the changelog for 0.4?

src/main/asciidoc/index.adoc
```diff
@@ -33,6 +33,7 @@ The library consists of these modules:
 *belgif-rest-problem:*
 
 * Removed deprecated InvalidParamProblem: use InputValidationProblem, which supports both the legacy invalidParams[] and the new issues[] structure
+* Added requireIfPresent check for validating input(s) that must be present when a given target input is present
```
clone1612, line 36: Probably best to add it to a *belgif-rest-problem-validator* section as that's the module with the change

PR #73, opened 2024-06-29
Handle Spring Boot 3 NoResourceFoundException

### LINKED ISSUE #72: [spring] Map NoResourceFoundException
When navigating to a URL that isn't mapped to a RestController, a NoResourceFoundException is thrown.

This exception currently results in a internalServerError problem type (default mapping); but should be a 404 resourceNotFound one.

jpraet:
Fixes #72.

Furthermore, this PR introduces the following modules:
* belgif-rest-problem-spring-boot-2: targets Spring Boot 2.x
* belgif-rest-problem-spring-boot-3: targets Spring Boot 3.x

belgif-rest-problem-spring contains the classes that are common to Spring Boot 2.x and 3.x.

### DIFF
--- a/belgif-rest-problem-it/belgif-rest-problem-spring-boot-3-it/src/test/java/io/github/belgif/rest/problem/RestProblemSpringBoot3IT.java
+++ b/belgif-rest-problem-it/belgif-rest-problem-spring-boot-3-it/src/test/java/io/github/belgif/rest/problem/RestProblemSpringBoot3IT.java
@@ -1,8 +1,11 @@
 package io.github.belgif.rest.problem;
 
+import static org.hamcrest.Matchers.*;
+
 import java.util.Arrays;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.Test;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.web.server.LocalServerPort;
 
@@ -23,6 +26,16 @@ class RestProblemSpringBoot3IT extends AbstractRestProblemSpringBootIT {
     @Override
     protected Stream<String> getClients() {
         return Arrays.stream(Client.values()).map(Client::name);
+    }
+
+    // On Spring Boot 3.x only (for now), a proper resourceNotFound problem is returned
+    @Override
+    @Test
+    void notFound() {
+        getSpec().when().get("/not/found").then().assertThat()
+                .statusCode(404)
+                .body("type", equalTo("urn:problem-type:belgif:resourceNotFound"))
+                .body("detail", equalTo("No resource frontend/not/found found"));
     }
 
 }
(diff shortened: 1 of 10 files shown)


belgif-rest-problem-it/belgif-rest-problem-spring-boot-3-it/src/test/java/io/github/belgif/rest/problem/RestProblemSpringBoot3IT.java
```diff
@@ -26,10 +25,4 @@ protected Stream<String> getClients() {
         return Arrays.stream(Client.values()).map(Client::name);
     }
 
-    @Override
-    @Disabled("Throws org.springframework.web.servlet.resource.NoResourceFoundException on latest Spring Boot 3")
-    void notFound() {
-        super.notFound();
```
clone1612, line 32: This root test only verifies the response has HTTP status 404, it doesn't verify a Problem type with appropriate content was returned.

We should still override it here with additional assertions on the Problem response.

src/main/asciidoc/index.adoc
```diff
@@ -342,6 +354,18 @@ This handles integration with the https://docs.spring.io/spring-framework/refere
 
 In general, these components make it possible to use standard java exception handling (throw and try-catch) for dealing with problems in Spring Boot REST APIs.
 
+[[belgif-rest-problem-spring-boot-2]]
+=== belgif-rest-problem-spring-boot-2
+
+Rather than depending on `belgif-rest-problem-spring` directly, Spring Boot 2.x users are recommended to depend on `belgif-rest-problem-spring-boot-2`, which adds some Spring Boot 2.x specific integrations.
```
clone1612, line 360: Should we change `belgif-rest-problem-spring` to a section reference? Now we have to scroll up to the start when we want to see its actual features.

clone1612 (changes requested):

belgif-rest-problem-it/belgif-rest-problem-spring-boot-3-it/src/test/java/io/github/belgif/rest/problem/RestProblemSpringBoot3IT.java
```diff
@@ -26,10 +28,14 @@ protected Stream<String> getClients() {
...
+    // On Spring Boot 3.x only (for now), a proper resourceNotFound problem is returned
     @Override
-    @Disabled("Throws org.springframework.web.servlet.resource.NoResourceFoundException on latest Spring Boot 3")
+    @Test
     void notFound() {
-        super.notFound();
+        getSpec().when().get("/not/found").then().assertThat()
+                .statusCode(404)
+                .body("type", equalTo("urn:problem-type:belgif:resourceNotFound"))
+                .body("detail", equalTo("No resource frontend/not/found found"));
```
clone1612, line 38: Interesting that the leading `/` is removed by Spring, should we maybe add it if the path doesn't start with one? Examples in Belgif guide have a leading `/` for their resource path.

Otherwise we should probably update our unit test as it currently doesn't reflect the actual exception thrown by Spring.

clone1612 (changes requested):

clone1612 (approved):

Issue #77, opened 2024-07-02
Disallow inputs array (inputvalidationissue) of size one


usku01:
By convention, the inputs[] array in InputValidationIssue should only be used when there are multiple referenced inputs.
For referencing a single input, in/name/value should be used instead.

This behavior should be enforced by the rest-problem library when adding input validation issues via the available methods.

PR #78, opened 2024-07-08
Disallow size one input list in InputValidationIssue

usku01:
The goal of this PR is to solve #77 

The approach is to evaluate the setter methods as pure setters and the 'add' methods as methods to add input (or in, name, value) to what already exist.


belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -133,34 +143,84 @@ public List<Input<?>> getInputs() {
...
-        verifyNoInNameValue();
+
         this.inputs.clear();
-        this.inputs.addAll(inputs);
+
+        if (inputs == null) {
+            return;
+        }
+
+        if (in != null || name != null || value != null) {
```
jpraet, line 153: hasInNameValue()

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -133,34 +143,84 @@ public List<Input<?>> getInputs() {
...
+            this.inputs.add(input);
+        } else {
+            in(input.getIn(), input.getName(), input.getValue());
+        }
     }
 
-    private void verifyNoInputs() {
-        if (!inputs.isEmpty()) {
-            throw new IllegalArgumentException(INPUTS_AND_IN_NAME_VALUE_ARE_MUTUALLY_EXCLUSIVE);
+    public void addInput(List<Input<?>> input) {
```
jpraet, line 196: addInput**s**?

jpraet (changes requested):

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -133,34 +143,84 @@ public List<Input<?>> getInputs() {
...
+            this.inputs.add(input);
+        } else {
+            in(input.getIn(), input.getName(), input.getValue());
+        }
     }
 
-    private void verifyNoInputs() {
-        if (!inputs.isEmpty()) {
-            throw new IllegalArgumentException(INPUTS_AND_IN_NAME_VALUE_ARE_MUTUALLY_EXCLUSIVE);
+    public void addInputs(List<Input<?>> input) {
```
jpraet, line 196: Are we giving the user too many options?

setInputs, addInput, addInputs, input, inputs

jpraet (commented):
@clone1612 could you also take a look

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
clone1612: Yes, I would streamline them:
* inputs (list and varargs)
  * I would drop input as it suggests a single one could be enough to create a valid instance (which we'll block in this PR)
  * "These are the inputs"
* addInputs (list and varags)
  * Drop addInput for same reason as input 

I would simply drop setInputs, its use case is covered inputs

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -133,34 +143,84 @@ public List<Input<?>> getInputs() {
...
-        if (in != null || name != null || value != null) {
+    public void addInput(Input<?>... input) {
+        if (input == null) {
+            return;
+        }
+        addInputs(Arrays.asList(input));
+    }
+
+    private void verifyNoInputs(Object valueToUpdate) {
+        if (valueToUpdate == null) {
```
clone1612, line 211: I wonder if this check with skip won't ignore undesired behavior. 

The method is called for any of the in/name/value setters. For in and name it doesn't make sense to be `null` to we'll ignore those. That leaves us with `setValue(null)`. If inputs wasn't empty we're attempting to combine 2 different concepts, even though the value is null. Shouldn't we continue to block that?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
usku01: >   * Drop addInput for same reason as input
> 

I don't know about deleting addInput. It kind of means that you want to add an input to what  alreadt exists (even if there is nothing). But adding a single input is covered by addInput(varargs) in any case

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
usku01: To me, the goal of adding this ==null was about, if the setting parameter=null, then there is no reason to check if there is already anything in inputs list because we are not adding information. In my opinion, it would almost be a bug if we did this check in that case.

But, about this statement:

>  For in and name it doesn't make sense to be null to we'll ignore those

It would not make sense indeed but the code does not check if these are null but value is not for instance. So it would be interesting to add this validation. What do you guys think? Or it should be "free" like in the open-api contracts?

To make things simpler, we could just provide the in(in, name, value) method in which we would check first if the combination of provided info is acceptable. all the other methods would be private (in(), name(), etc).

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssueTest.java
```diff
@@ -129,45 +152,160 @@ void inNameValue() {
     @Test
     void inNameValueAndInputsAreMutuallyExclusive() {
         Input<?> input = Input.query("name", "value");
+        List<Input<?>> inputs = Arrays.asList(input, input);
+
+        assertThatIllegalArgumentException().isThrownBy(
```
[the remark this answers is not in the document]
clone1612: What about defining a utility function to reduce some of the duplication?

```
    private void assertThatMutuallyExclusiveExceptionIsThrownBy(ThrowableAssert.ThrowingCallable throwingCallable) {
        assertThatIllegalArgumentException()
                .isThrownBy(throwingCallable)
                .withMessageContaining(MUTUALLY_EXCLUSIVE_EXC);
    }
```

Your test assertions then become a bit simpler and more readable:

```
        assertThatMutuallyExclusiveExceptionIsThrownBy(() -> new InputValidationIssue()
                .inputs(input, input)
                .value("value"));
```

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
clone1612: It's hard to imagine all ways a developer would use it, but maybe it can be simply become
* inputs (collection and varargs)
  * These are the input for this issue
* addInput(Input input)
* addInputs(collection)
  * No varargs as it doesn't seem that useful, you add a single entry or multiple you've collected elsewhere (and most likely into a collection

I wouldn't even be opposed against dropping "addInputs(collection)" and simply supporting:
* inputs (collection and varargs)
* addInput(Input input)

I feel that is simple enough while still supporting all use-cases.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
clone1612: Ok to keep this check.

> It would not make sense indeed but the code does not check if these are null but value is not for instance. So it would be interesting to add this validation. What do you guys think?

I don't think it's possible to add a check for that since we don't have "build" method that indicates we're done with the initialization. When seeing a value but no in/name we don't know it they'll be set later on by the developer.

Individual "in()", "name()", ... methods make construction a bit more readable, especially if another method is called to determine their value.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
clone1612: As discussed during the meeting we'll go ahead with the last suggestion:
- inputs (collection and varargs) for fluent building of the issue
- addInput(Input input) for dynamically adding input, e.g. in a for-loop

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -68,6 +72,12 @@ public InputValidationIssue(InEnum in, String name) {
         this.name = name;
     }
 
+    public InputValidationIssue(Input<?> input) {
```
jpraet, line 75: I don't think we really need this new constructor? It's only used in a test.

I'm not too happy with the pre-existing "in, name (, value)" constructors either, but probably not worth it to remove those. With these constructors, we facilitate the bad practice of not populating the type and title.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -288,18 +307,42 @@ public InputValidationIssue valueEntry(String key, Object value) {
...
+        if (filteredInputs.size() == 1) {
+            throw new IllegalArgumentException(INPUTS_SETTER_ONE_ITEM);
+        }
+
+        this.inputs.clear();
+        this.inputs.addAll(filteredInputs);
         return this;
     }
 
     public InputValidationIssue inputs(Input<?>... inputs) {
```
jpraet, line 340: As we want to enforce at least two inputs, maybe we could change the method signature to

`inputs(Input<?> first, Input<?> second, Input<?>... otherInputs)`

which will catch the error at compile time rather than runtime?

Do you see any drawbacks with that @usku01 @clone1612?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -124,41 +134,49 @@ public Object getValue() {
...
-        verifyNoInputs();
+        verifyNoInputs(value);
         this.value = value;
     }
 
     public List<Input<?>> getInputs() {
         return Collections.unmodifiableList(inputs);
     }
 
-    public void setInputs(List<Input<?>> inputs) {
```
jpraet, line 135: Should we remove them now, or first mark them `@Deprecated(forRemoval = true)` and only remove them when going to 1.0?

The inputs[] is not belgif-standard, it's a CBSS extension.
So I guess the risk is low that anyone outside of CBSS would be impacted by this removal, and removing them immediately should not be too problematic?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
usku01: > Do you see any drawbacks with that @usku01 @clone1612?

I'm ok with this proposal.
But note that it will not necessarily prevent the INPUTS_SETTER_ONE_ITEM error from happening (I don't know if it is part of your goal). Depending on how the provided inputs are generated, the second input could be null for instance (and no other inputs are provided).

clone1612: Also fine for me

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
usku01: > Should we remove them now, or first mark them `@Deprecated(forRemoval = true)` and only remove them when going to 1.0?

It does not seem to be used by any project (I did not check the PR though) so I don't think it's a problem to remove it.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
usku01: > I don't think we really need this new constructor? It's only used in a test.
 Ok to remove it

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -288,18 +307,42 @@ public InputValidationIssue valueEntry(String key, Object value) {
...
+    /**
+     *
+     * @param inputs list with input items to initialize the inputs parameter. Any previous inputs are erased from the
+     *        parameter
+     * @return this
+     *
+     * @throws IllegalArgumentException if in/name/value parameters are not null (because it's mutually exclusive). This
+     *         exception is also thrown if the list contains only one non-null item.
+     */
     public InputValidationIssue inputs(List<Input<?>> inputs) {
```
jpraet, line 319: As Jannick noted during our meeting, it can accept a `Collection<Input<?>>`.

jpraet (approved):

jpraet:
OK for me. @clone1612 ?

clone1612 (approved):

Issue #98, opened 2024-09-20
ProblemWebClientCustomizer not compatible with Spring Boot 3


pvdbosch:
We got following exception when testing with spring boot 3 webclient :

```
java.lang.NoSuchMethodError: 'org.springframework.http.HttpStatus org.springframework.web.reactive.function.client.ClientResponse.statusCode()'
	at io.github.belgif.rest.problem.spring.ProblemWebClientCustomizer.lambda$static$0(ProblemWebClientCustomizer.java:24) ~[belgif-rest-problem-spring-0.6.0-jakarta.jar:na]

...
```

Apparently [ClientResponse.statusCode()](https://github.com/spring-projects/spring-framework/blob/main/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientResponse.java#L58) method signature changed in Spring Framework 6 from HttpStatus to HttpStatusCode, causing this error.
In Spring 6, HttpStatus has been retrofitted to implement HttpStatusCode, but even so, this seems to cause compatibility issues in  ProblemWebClientCustomizer which is compiled against Spring 5.

jpraet:
It's a bit surprising because we have an integration test for it:

https://github.com/belgif/rest-problem-java/blob/e8f435d0ff4b11b291a12bf57507f3c85cd7b170/belgif-rest-problem-it/belgif-rest-problem-spring-boot-3-it/src/main/java/io/github/belgif/rest/problem/FrontendController.java#L103-L104

What spring boot version do you use?

smals-mavh:
@jpraet @pvdbosch ,

We used 3.3.3, however, we had the same result with SB 3.3.1.

I've added a reproducer https://github.com/belgif/rest-problem-java/pull/100

PR #100, opened 2024-09-23
Reproducer for issue added in IT

smals-mavh:
Reproducer for https://github.com/belgif/rest-problem-java/issues/98
Because:

https://github.com/belgif/rest-problem-java/blob/e8f435d0ff4b11b291a12bf57507f3c85cd7b170/belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/ProblemWebClientCustomizer.java#L23-L24

The first statement already returns true in all IT's, the statement to check statuscode is not checked in ITs.

Added reproducer for a happyPath. (does not fix the issue, build doesn't work, just a reproducer).


jpraet:
Thanks for the reproducer, integrated into #102 which has the fix.

PR #101, opened 2024-09-23
Extensible validator

jpraet:
First attempt for extensible validator, based on http://egalluzzo.blogspot.com/2010/06/using-inheritance-with-fluent.html.

But I'm not very happy with it because the inheritance may be too limiting.
e.g. if you have a validator for business domain A and a validator for business domain B, and you have an application that deals with both A and B, you would want to be able to combine these into one validator. That is not currently possible without having an inheritance relationship between A and B.

I will look into making it more composable. I think it should be doable with default methods on interfaces.


belgif-rest-problem-validator/src/main/java/io/github/belgif/rest/problem/validation/AbstractRequestValidator.java
```diff
@@ -0,0 +1,447 @@
...
+     * @return this RequestValidator
+     */
+    public V custom(InputValidator validator) {
+        if (validator != null) {
+            addValidator(validator);
+        }
+        return getThis();
+    }
+
+    protected abstract V getThis();
```
jflabatBCSS, line 441: Don't you do like assertj by giving the `Class<?> selfType` in the constructor and store the self type by doing `myself = (SELF) selfType.cast(this)`. It requires a cast in the abstract constructor but it's easier to use by the class extending it.

jpraet: Something like this then https://github.com/belgif/rest-problem-java/pull/101/commits/a17cf50c6df559daef7ac42331f00eb4bc506d8d?

jflabatBCSS (approved):

jflabatBCSS (approved):

jpraet:
@pvdbosch does this look OK to you, or do you see the need to go further with #103?

pvdbosch:
OK, single inheritance seems sufficient to me for most use cases

PR #103, opened 2024-09-23
Extensible and modular request validator

jpraet:
Alternate approach to #101.

It seems a bit convoluted with all the self-referencing generics. I'm open to other ideas.


jflabatBCSS:
> Alternate approach to #101.
> 
> It seems a bit convoluted with all the self-referencing generics. I'm open to other ideas.

Indeed. I think in practice it makes it difficult to use but is it frequent use case to have different domains that you would combine together. 

It's also abusing the interfaces to make a class extending two parents.

jpraet:
I had in mind that the standard belgif validations could perhaps also be split up into modules like SsinValidatorModule, EnterpriseNumberValidatorModule, CrossParameterValidatorModule, ...

Because Smals has already let us know that they use a different bean validation approach for validating ssins, enterprise numbers, ... so they could then decide to create their own validator that does not include those modules.

But I don't know, they could also just ignore those methods and not use them :wink:.

Whether it's really abusing interfaces or not is up for debate :smile:.
It's kind of a well established (anti-?)pattern to use interfaces in this way to have mixins in java.

jflabatBCSS:
> I had in mind that the standard belgif validations could perhaps also be split up into modules like SsinValidatorModule, EnterpriseNumberValidatorModule, CrossParameterValidatorModule, ...
> 
> Because Smals has already let us know that they use a different bean validation approach for validating ssins, enterprise numbers, ... so they could then decide to compose their own validator that does not include those modules.
> 
> But I don't know, they could also just ignore those methods and not use them 😉.
> 
> Whether it's really abusing interfaces or not is up for debate 😄. It's kind of a well established (anti-?)pattern to use interfaces in this way to have mixins in java.

Ok. But we have to choose between both implementations. If Smals already uses different modules, maybe this solution is the way to go.

jflabatBCSS (approved):

jflabatBCSS (approved):

jpraet:
Closing in favor of #101.

PR #105, opened 2024-09-25
Retrieve bean validation parameter names from Spring MVC / JAX-RS annotations

### LINKED ISSUE #96: Bug: parameter name not correctly retrieved from interface
Got the following response from a Spring Boot 3.3.3 application:
```
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "detail": "must be greater than or equal to 197",
            "in": "path",
            "name": "arg0",
            "value": 160
        }
    ]
}
```
Parameter annotations are only present on the (generated) interface, not on the implementation (when adding annotation to the implementation, 'name' is filled in correctly). Probably something goes wrong in reading the annotations of the parent class / interface.

jpraet:
Fixes #96


belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/jaxrs/JaxRsParameterNameProvider.java
```diff
@@ -0,0 +1,70 @@
...
+        return parameterNameCache.computeIfAbsent(constructor, this::getParameterNames);
+    }
+
+    @Override
+    public List<String> getParameterNames(Method method) {
+        return parameterNameCache.computeIfAbsent(method, this::getParameterNames);
+    }
+
+    private List<String> getParameterNames(Executable executable) {
+        Parameter[] parameters = executable.getParameters();
```
clone1612, line 42: Wouldn't a stream be more compact?

```
        return Arrays.stream(executable.getParameters())
                .map(this::getParameterName)
                .collect(Collectors.toList());
```

belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/jaxrs/JaxRsParameterNameProviderTest.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class JaxRsParameterNameProviderTest {
+
+    private JaxRsParameterNameProvider provider = new JaxRsParameterNameProvider();
```
clone1612, line 18: Field may be final

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/AnnotationParameterNameDiscoverer.java
```diff
@@ -0,0 +1,61 @@
...
+        return parameterNameCache.computeIfAbsent(constructor, this::getParameterNames);
+    }
+
+    @Override
+    public String[] getParameterNames(Method method) {
+        return parameterNameCache.computeIfAbsent(method, this::getParameterNames);
+    }
+
+    private String[] getParameterNames(Executable executable) {
+        Parameter[] parameters = executable.getParameters();
```
clone1612, line 35: Wouldn't a stream be a bit more compact?

```
        return Arrays.stream(executable.getParameters())
                .map(this::getParameterName)
                .toArray(String[]::new);
```

belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/AnnotationParameterNameDiscovererTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.params.provider.ValueSource;
+import org.springframework.web.bind.annotation.CookieValue;
+import org.springframework.web.bind.annotation.MatrixVariable;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestParam;
+
+class AnnotationParameterNameDiscovererTest {
+
+    private AnnotationParameterNameDiscoverer discoverer = new AnnotationParameterNameDiscoverer();
```
clone1612, line 16: Field can be made final

belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/ProblemValidatorConfigurationTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.mockito.Mockito;
+import org.springframework.context.ApplicationContext;
+import org.springframework.validation.beanvalidation.LocalValidatorFactoryBean;
+
+class ProblemValidatorConfigurationTest {
+
+    private ProblemValidatorConfiguration config = new ProblemValidatorConfiguration();
```
clone1612, line 13: Field can be made final

belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/ProblemValidatorConfigurationTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Test
+    void defaultValidator() throws Exception {
+        LocalValidatorFactoryBean validator = config.defaultValidator(Mockito.mock(ApplicationContext.class));
+        Object parameterNameDiscoverer = ReflectionUtils
```
clone1612, line 18: No need to use any reflection utils, AssertJ can already handle this scenario:

```
        assertThat(validator)
                .extracting("parameterNameDiscoverer")
                .isInstanceOf(AnnotationParameterNameDiscoverer.class);
```

clone1612 (changes requested):
Additional question: do we keep using `<maven.compiler.parameters>true</maven.compiler.parameters>`?

jpraet:
> Additional question: do we keep using `<maven.compiler.parameters>true</maven.compiler.parameters>`?

I prefer to keep it, yes. I think it's a good practice. I wonder why it's not enabled by default actually.

When there is no annotation, the ParameterNameProvider/Discoverer falls back to the parameter name.

Also: for the Spring MVC annotations, the name in the annotation is actually optional. So there, we should also fall back to the parameter name, which I overlooked in the current implementation.

belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/jaxrs/JaxRsParameterNameProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A ParameterNameProvider that retrieves the parameter name from JAX-RS annotations (if present).
+ *
+ * @see ParameterNameProvider
+ */
+@Provider
+public class JaxRsParameterNameProvider implements ParameterNameProvider {
+
+    private final ConcurrentHashMap<Executable, List<String>> parameterNameCache = new ConcurrentHashMap<>();
```
jpraet, line 29: Maybe we should make this cache static? Because I don't know if the JAX-RS / Bean Validation spec defines that there will only be a single reused instance of the ParameterNameProvider.

clone1612: The interface documentation mentions that the implementation should be thread-safe, so I assume it would be reused? It wouldn't hurt to make it static though, that way we're certain.

clone1612:
> > Additional question: do we keep using `<maven.compiler.parameters>true</maven.compiler.parameters>`?
> 
> I prefer to keep it, yes. I think it's a good practice. I wonder why it's not enabled by default actually.
> 
> When there is no annotation, the ParameterNameProvider/Discoverer falls back to the parameter name.

Ok, but wouldn't it be better to then document that best practice in the documentation?

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/AnnotationParameterNameDiscoverer.java
```diff
@@ -0,0 +1,63 @@
...
+                .map(this::getParameterName)
+                .toArray(String[]::new);
+    }
+
+    private String getParameterName(Parameter parameter) {
+        Annotation[] annotations = parameter.getAnnotations();
+        String parameterName = null;
+        for (Annotation annotation : annotations) {
+            if (annotation instanceof RequestParam) {
+                parameterName = ((RequestParam) annotation).value();
```
clone1612, line 46: Shouldn't we break upon encountering a parameter annotation? Now we'll traverse all annotations on the parameter, including all the validations ones (if present).

Or not worth the uglier code as result will be cached?

belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/AnnotationParameterNameDiscovererTest.java
```diff
@@ -0,0 +1,66 @@
...
+        void requestHeader(@RequestHeader("name") String x) {
+        }
+
+        void cookieValue(@CookieValue("name") String x) {
+        }
+
+        void matrixVariable(@MatrixVariable("name") String x) {
+        }
+
+        void fallback(@RequestParam String name) {
```
clone1612, line 61: I would add an equivalent integration test for this to verify the fallback works in an actual application

clone1612 (changes requested):

clone1612 (approved):

jpraet:
wrong button

PR #107, opened 2024-09-27
Add Spring ExceptionHandler for HTTP 405, 406 and 415

### LINKED ISSUE #38: Wrong Request method returns a 500 internal server error (Spring Boot)
When calling an endpoint with the wrong request method, Spring boot throws a HttpRequestMethodNotSupportedException.
Currently this exception is not handled.

jpraet:
Fixes #38.

### DIFF
--- a/belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/RoutingExceptionsHandlerTest.java
+++ b/belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/RoutingExceptionsHandlerTest.java
@@ -4,13 +4,20 @@ import static io.github.belgif.rest.problem.api.InputValidationIssues.*;
 import static org.assertj.core.api.Assertions.*;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Objects;
 
 import org.junit.jupiter.api.Test;
 import org.springframework.core.MethodParameter;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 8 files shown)


belgif-rest-problem-spring/src/test/java/io/github/belgif/rest/problem/spring/RoutingExceptionsHandlerTest.java
```diff
@@ -71,4 +78,27 @@ void handleHttpMessageNotReadable() {
...
+    @Test
+    void handleHttpMediaTypeNotAcceptable() {
+        ResponseEntity<Void> response = handler.handleHttpMediaTypeNotAcceptable(
+                new HttpMediaTypeNotAcceptableException(Collections.singletonList(MediaType.APPLICATION_JSON)));
+        assertThat(response.getStatusCodeValue()).isEqualTo(406);
+    }
+
+    @Test
+    void handleHttpMediaTypeNotSupported() {
+        ResponseEntity<Void> response = handler.handleHttpMediaTypeNotSupported(
```
jflabatBCSS, line 98: I don't understand. If read Peter, the specs state we should include the supported media type. We are not including because it's difficult to retrieve them in spring. But here, I see it's possible to get them from the exception. Should be not at least support if they are present ?

jpraet: Noted in https://github.com/belgif/rest-problem-java/issues/38#issuecomment-2379094962.

But in any case, the problem type and how to represent the supported media types in the problem body has not been standardized yet. So for now, a HTTP 406 with empty body should suffice (better than the current HTTP 500).

jflabatBCSS (approved):

PR #109, opened 2024-09-28
Bypass ProblemClientResponseFilter on MicroProfile REST Client

### LINKED ISSUE #31: MicroProfile Client ProblemResponseExceptionMapper is not called
As observed by the code coverage, it turns out the [ProblemResponseExceptionMapper](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/jaxrs/client/ProblemResponseExceptionMapper.java) is never called. Instead, the JAX-RS [ProblemClientResponseFilter](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/jaxrs/client/ProblemClientResponseFilter.java) takes precedence, and the thrown ResponseProcessingException is being unwrapped by:

https://github.com/resteasy/resteasy/blob/120abda3249de866a8269b9fd2128ddce318eedb/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java#L157-L162.

Can we remove the ProblemResponseExceptionMapper? 
Are all MicroProfile REST client implementations guaranteed to implement this unwrapping behavior (which differs from JAX-RS)?
Or alternatively can we prevent the ProblemClientResponseFilter from being picked up by MicroProfile REST client?

jpraet:
Fixes #31.

Bypass ProblemClientResponseFilter on MicroProfile REST Client (by checking for 
[org.eclipse.microprofile.rest.client.invokedMethod](https://download.eclipse.org/microprofile/microprofile-rest-client-4.0/microprofile-rest-client-spec-4.0.html#_clientrequestfilter) property), so the MicroProfile-specific ProblemResponseExceptionMapper can be used instead.

### DIFF
--- a/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/jaxrs/client/ProblemClientResponseFilterTest.java
+++ b/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/jaxrs/client/ProblemClientResponseFilterTest.java
@@ -5,6 +5,7 @@ import static org.mockito.Mockito.*;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 
@@ -95,6 +96,15 @@ class ProblemClientResponseFilterTest {
         when(responseContext.getStatus()).thenReturn(400);
         assertThatNoException().isThrownBy(
                 () -> filter.filter(requestContext, responseContext));
+    }
+
+    @Test
+    void microProfile() {
+        when(requestContext.getProperty("org.eclipse.microprofile.rest.client.invokedMethod"))
+                .thenReturn(mock(Method.class));
+        assertThatNoException().isThrownBy(
+                () -> filter.filter(requestContext, responseContext));
+        verifyNoMoreInteractions(requestContext, responseContext);
     }
 
 }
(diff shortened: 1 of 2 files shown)


belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/jaxrs/client/ProblemClientResponseFilterTest.java
```diff
@@ -97,4 +98,13 @@ void differentMediaType() {
                 () -> filter.filter(requestContext, responseContext));
     }
 
+    @Test
+    void microProfile() {
+        when(requestContext.getProperty("org.eclipse.microprofile.rest.client.invokedMethod"))
```
jflabatBCSS, line 103: Is there a way to test that in the integration test ? My fear is that when the property value changes, you will go back to the old behaviour.

jpraet: The MicroProfile REST Client spec defines that this property MUST be set: https://download.eclipse.org/microprofile/microprofile-rest-client-4.0/microprofile-rest-client-spec-4.0.html#_clientrequestfilter.

I did verify it manually (looking at the code coverage for the integration test).

jflabatBCSS (approved):

PR #118, opened 2024-10-14
Add support for Atlassian swagger-request-validator library

smals-mavh:
Currently only implemented for Spring Boot 3.

atlassian lib for SB3 is different from the SB2 one ([swagger-request-validator-springmvc](https://bitbucket.org/atlassian/swagger-request-validator/src/master/swagger-request-validator-springmvc/) for sb2, [swagger-request-validator-spring-webmvc](https://bitbucket.org/atlassian/swagger-request-validator/src/master/swagger-request-validator-spring-webmvc/) for sb3).

Now the belgif-rest-problem-openapi-validation-spring library also comes with the belgif-rest-problem-spring-boot-3 lib which might be confusing. 

Maybe we should rename belgif-rest-problem-openapi-validation-spring to belgif-rest-problem-openapi-validation-spring-boot-3 to be more clear on the fact that it (now) only supports SB3, and it already includes the other errormappers.

Or if someone has a better / more efficient idea, please let me know.

@jpraet ready for review.


belgif-rest-problem-openapi-validation-spring/src/main/java/io/github/belgif/rest/problem/openapi/OpenapiRequestViolationProblemAdvice.java
```diff
@@ -0,0 +1,181 @@
...
+import io.github.belgif.rest.problem.ResourceNotFoundProblem;
+import io.github.belgif.rest.problem.api.InEnum;
+import io.github.belgif.rest.problem.api.InputValidationIssue;
+import io.github.belgif.rest.problem.api.Problem;
+import io.github.belgif.rest.problem.spring.ProblemMediaType;
+import io.swagger.v3.oas.models.parameters.Parameter;
+
+@RestControllerAdvice
+@ConditionalOnWebApplication
+public class OpenapiRequestViolationProblemAdvice {
```
jpraet, line 38: I wonder if it can't be added directly to belgif-rest-problem-spring, with
@ConditionalOnClass(InvalidRequestException.class)?

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-it/src/main/java/io/github/belgif/rest/problem/OpenApiValidationConfig.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Bean
+    public WebMvcConfigurer addOpenApiValidationInterceptor(@Value("/openapi.yaml") final String apiSpecification,
+            @Value("server.servlet.context-path") String contextPath) {
+        final OpenApiInteractionValidator validator = OpenApiInteractionValidator
+                .createForSpecificationUrl(apiSpecification)
+                .withLevelResolver(LevelResolver.create()
+                        // Accept unknown query parameters
+                        .withLevel("validation.request.parameter.query.unexpected", ValidationReport.Level.IGNORE)
```
pvdbosch, line 33: I think this check could be decided per-API what the desired behavior would be.
Validation behavior isn't really specified in the OpenAPI standard in case of unknown query parameters.

In case it isn't ignored, I think we should return a badRequest problem with a `urn:problem-type:belgif:input-validation:unknownInput` issue type [[req-valid]](https://www.belgif.be/specification/rest/api-guide/#rule-req-valid).

belgif-rest-problem-openapi-validation-spring/src/main/java/io/github/belgif/rest/problem/openapi/OpenapiRequestViolationProblemAdvice.java
pvdbosch, line 1: for consistency: Open**A**piRequestViolationProblemAdvice

pvdbosch (changes requested):

pvdbosch:
@smals-mavh , would be best to update doc for this new feature, but might have to be coordinated with other doc changes in https://github.com/belgif/rest-problem-java/tree/feature/%23117-improve-documentation

belgif-rest-problem-openapi-validation-spring/src/main/java/io/github/belgif/rest/problem/openapi/OpenapiRequestViolationProblemAdvice.java
smals-mavh: I'll try

smals-mavh: Seems to work fine.
Only difference between SB2 and 3 is that when a non-existing path is called, in SB3, this can be validated by the atlassian lib and catched by the controlleradvice, where in sb2, the exception is already thrown in the dispatcherServlet, and cannot be catched by the controlleradvice. When the atlassian lib is configured to allow unknown paths, spring handles the 404.

belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/AbstractOpenApiValidationSpringBootIT.java
```diff
@@ -0,0 +1,172 @@
...
+        getSpec().when().get("/myFirstPath?myUnknownParam=123").then().assertThat()
+                .statusCode(400).log().all()
+                .body("type", equalTo("urn:problem-type:belgif:badRequest"))
+                .body("issues[0].type", equalTo("urn:problem-type:belgif:input-validation:unknownInput"))
+                .body("issues[0].title", equalTo("Unknown query parameter."))
+                .body("issues[0].in", equalTo("query"))
+                .body("issues[0].name", equalTo("myUnknownParam"))
+                .body("issues[0].value", equalTo("123"))
+                .body("issues[0].detail",
+                        equalTo("Query parameter 'myUnknownParam' is unexpected on path \"/myFirstPath\""));
```
pvdbosch, line 31: For all detail and title fields: avoid testing on an exact value within a free-text description.
It makes the tests fragile: these descriptions might be changed freely, especially when created by a 3rd party library. It's not really a regression/bug when the wording of the description changes.

Checking that the string is non-empty may suffice.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java
```diff
@@ -0,0 +1,198 @@
...
+import io.github.belgif.rest.problem.api.InEnum;
+import io.github.belgif.rest.problem.api.InputValidationIssue;
+import io.github.belgif.rest.problem.api.Problem;
+import io.swagger.v3.oas.models.parameters.Parameter;
+
+@RestControllerAdvice
+@ConditionalOnWebApplication
+@Order(1)
+@ConditionalOnClass(InvalidRequestException.class)
+public class OpenApiRequestViolationProblemAdvice {
```
pvdbosch, line 41: Can you add javadoc on the class to explain its purpose? 
Also add comment why `@Order(1)`

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-3-it/src/test/java/io/github/belgif/rest/problem/RestProblemOpenapiValidationSpringBoot3IT.java
```diff
@@ -0,0 +1,32 @@
...
+    @LocalServerPort
+    private int port;
+
+    @Override
+    protected RequestSpecification getSpec() {
+        return RestAssured.with().baseUri("http://localhost").port(port).basePath("/spring");
+    }
+
+    // Only seems to work in SB3. In SB2 the atlassian lib throws the same exception, but the missing path seems not to
+    // be handled by a controller and not catched by the ControllerAdvice.
```
pvdbosch, line 24: I think you can add this test to superclass and override it adding @Ignore to the SB2 subclass?

pvdbosch (changes requested):

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private List<InputValidationIssue> constructIssues(InvalidRequestException ex, HttpServletRequest request) {
+        Set<InputValidationIssue> issues = new HashSet<>();
+        for (ValidationReport.Message message : ex.getValidationReport().getMessages()) {
+            InputValidationIssue issue = handleSpecialCases(message);
+            if (issue != null) {
+                issues.add(issue);
+                continue;
+            }
+            issue = getBasicIssue(message);
```
jpraet, line 72: Can you use the existing factory method for constructing the InputValidationIssue?

https://github.com/belgif/rest-problem-java/blob/9cfda64728f64775b2765ddd87dbaeba1a36fe6c/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java#L61-L66

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-2-it/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>3.8.0</version>
```
jpraet, line 85: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-2-it/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+              <overWriteSnapshots>true</overWriteSnapshots>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.openapitools</groupId>
+        <artifactId>openapi-generator-maven-plugin</artifactId>
+        <version>7.9.0</version>
```
jpraet, line 113: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-2-it/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+              </configOptions>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-clean-plugin</artifactId>
+        <version>3.4.0</version>
```
jpraet, line 140: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-3-it/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>3.8.0</version>
```
jpraet, line 84: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-3-it/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+              <overWriteSnapshots>true</overWriteSnapshots>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.openapitools</groupId>
+        <artifactId>openapi-generator-maven-plugin</artifactId>
+        <version>7.9.0</version>
```
jpraet, line 112: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-it/belgif-rest-problem-openapi-validation-spring-boot-3-it/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+              </configOptions>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-clean-plugin</artifactId>
+        <version>3.4.0</version>
```
jpraet, line 139: Version can be removed (it's in the pluginManagement of https://github.com/belgif/rest-problem-java/blob/main/pom.xml).

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java
jpraet: Strange, I removed it locally to see if it was really needed, and the build succeeded.
But on the github build it fails. So I guess if multiple exception handlers have the same priority, the order is undetermined.

clone1612 (changes requested):

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * This class is only active when Spring-Boot applications implemented the Atlassian swagger-request-validator.
+ * Maps the InvalidRequestException thrown by the swagger-request-validator to belgif problems.
+ */
+@RestControllerAdvice
+@ConditionalOnWebApplication
+@ConditionalOnClass(InvalidRequestException.class)
+@Order(1)
+// @Order(1) to take precedence over io.github.belgif.rest.problem.spring.ProblemExceptionHandler
+public class OpenApiRequestViolationProblemAdvice {
```
jpraet, line 46: The naming pattern is different from the existing exception handlers:
* ProblemExceptionHandler
* RoutingExceptionsHandler
* BeanValidationExceptionsHandler
* NoResourceFoundExceptionHandler

(well... they are already a bit inconsistent by themselves with ExceptionHandler vs ExceptionsHandler).

belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/AbstractOpenApiValidationSpringBootIT.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private static final String BAD_REQUEST_URN = "urn:problem-type:belgif:badRequest";
+    private static final String SCHEMA_VIOLATION_URN = "urn:problem-type:belgif:input-validation:schemaViolation";
+
+    protected abstract RequestSpecification getSpec();
+
+    @Test
+    void validPathParamWorksTest() {
+        getSpec().when().get("/myFirstPath/abc1234567").then().assertThat()
+                .statusCode(200).log().all()
```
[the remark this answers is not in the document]
smals-mavh: Removed the log().all() to not clutter the logs to much.
Can easily be added back in if needed for debugging purposes.

jpraet:
@smals-mavh I have done some refactoring in #121.
Can you have a look at the changes, and merge them into this PR if you agree?

belgif-rest-problem-it/belgif-rest-problem-spring-boot-2-it/src/main/java/io/github/belgif/rest/problem/OpenApiValidationConfig.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public void addInterceptors(@NonNull InterceptorRegistry registry) {
+        OpenApiInteractionValidator validator = OpenApiInteractionValidator
+                .createForSpecificationUrl("/openapi.yaml")
+                .withLevelResolver(LevelResolver.create()
+                        // Accept additionalProperties even if they're not defined in the schema
+                        .withLevel("validation.schema.additionalProperties", ValidationReport.Level.IGNORE)
+                        // Ignores validation when a path is not in the openapi and let Spring handle the error
```
jpraet, line 36: This also means this code is not used:

https://github.com/belgif/rest-problem-java/blob/da8daa2ad77026bd8235d3751e061835cabd6997/belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java#L59-L61

As can be seen in the [code coverage](https://sonarcloud.io/component_measures?metric=new_coverage&selected=belgif_rest-problem-java%3Abelgif-rest-problem-spring%2Fsrc%2Fmain%2Fjava%2Fio%2Fgithub%2Fbelgif%2Frest%2Fproblem%2Fspring%2FOpenApiRequestViolationProblemAdvice.java&view=list&pullRequest=118&id=belgif_rest-problem-java).

I tested what happens if we remove this level customization, and indeed, the InvalidRequestException does NOT get sent to the `@ExceptionHandler(value = InvalidRequestException.class)`.

See also https://github.com/spring-projects/spring-framework/issues/22619#issuecomment-658661059.

The InvalidRequestException then gets converted to an HTTP 404 with a payload that is not problem-compliant.

So, if we can't figure out how to handle this case, then we should document that it is required to disable this `validation.request.path.missing` validation.

smals-mavh:
@clone1612 @jpraet @pvdbosch ,

Tried to implement all the feedback.

The atlassian library gives one validation (message in validation report) for each object that is missing a required property. When one object is missing multiple properties, it provides all missing properties in the detail message, and the 'name' is the parent object. I think this is clear enough, and it might not be worth our time to split this out even further, because it would involve regex on the message (which might change / be translated).
Example:

```
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "detail": "[Path '/myInnerObject'] Object has missing required properties ([\"myParam\"])",
            "in": "body",
            "name": "/myInnerObject"
        },
        {
            "detail": "Object has missing required properties ([\"myFirstProperty\"])",
            "in": "body",
            "name": "/"
        }
    ]
}
```
or
```
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "detail": "Object has missing required properties ([\"myFirstProperty\",\"myInnerObject\"])",
            "in": "body",
            "name": "/"
        }
}
```
The / in name is because it's in the root schema of the request body.

For the allOf you had a good catch @clone1612 !
Modified it to 'explode' the messages in case of allOf:
```
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "detail": "[Path '/myInnerObject/myParam'] String \"abc\" is too short (length: 3, required minimum: 10)",
            "in": "body",
            "name": "/myInnerObject/myParam",
            "value": "abc"
        },
        {
            "detail": "[Path '/myInnerObject/myOtherParam'] String \"abc\" is too short (length: 3, required minimum: 10)",
            "in": "body",
            "name": "/myInnerObject/myOtherParam",
            "value": "abc"
        }
    ]
}
```

But just getting all the nested messages wouldn't work for oneOf / anyOf schemas, so in those cases, the detail message will be different, but it will remain just one issue:
```
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "detail": "[Path '/myInnerObject'] Instance failed to match exactly one schema (matched 0 out of 2) :  -- [Path '/myInnerObject'] Object has missing required properties ([\"myParam\"]) -- [Path '/myInnerObject'] Object has missing required properties ([\"myOtherParam\"])",
            "in": "body",
            "name": "/myInnerObject"
        }
    ]
}
```

belgif-rest-problem-it/belgif-rest-problem-spring-boot-2-it/src/main/java/io/github/belgif/rest/problem/OpenApiValidationConfig.java
smals-mavh: This is only the case for SB2. 
I missed the fact that validation.request.path.missing was also disabled for sb3. (enabled it again)
That was why the test for the missing path is only disabled for SB2.

Was maybe a bit better described in the original codecomment: https://github.com/belgif/rest-problem-java/pull/118/files/3657fa3c5a92430a01110f97771df4c256c24f9f#diff-6d0d73833ab5b51decf31ceea63039d8ad16eb13250383001806d1dd40dc3dbe

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/OpenApiRequestViolationProblemAdvice.java
```diff
@@ -0,0 +1,180 @@
...
+        if (isNonExistingPath(ex)) {
+            return ProblemMediaType.INSTANCE.toResponse(new ResourceNotFoundProblem());
+        }
+        List<InputValidationIssue> issues = constructIssues(ex, request);
+        return ProblemMediaType.INSTANCE.toResponse(new BadRequestProblem(issues));
+    }
+
+    private List<InputValidationIssue> constructIssues(InvalidRequestException ex, HttpServletRequest request) {
+        Set<InputValidationIssue> issues = new HashSet<>();
+        for (ValidationReport.Message message : ex.getValidationReport().getMessages()) {
```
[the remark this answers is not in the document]
clone1612: @smals-mavh You're right it's probably not worth the effort/introduction of brittle logic. Ok for me to continue with this single issue approach.

clone1612 (approved):
@smals-mavh Changes look good, thanks! Don't forget to add a release note in `src/main/asciidoc/release-notes.adoc` when you're ready to merge.

smals-mavh:
Added documentation and releasenote (still for v0.9, since I didn't see that on maven central yet)

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,217 @@
...
+
+    private String getPathValue(ValidationReport.Message message, String name) {
+        String original = message.getContext().flatMap(ValidationReport.MessageContext::getApiOperation)
+                .map(p -> p.getApiPath().original()).orElse(null);
+        if (original != null) {
+            String actual = message.getContext()
+                    .flatMap(ValidationReport.MessageContext::getRequestPath)
+                    .orElse("");
+
+            String regex = original.replace("{" + name + "}", "(.*)").replaceAll("/\\{.*}", "/.*");
```
jpraet, line 148: I think there's a bug here if you have something like `/myFirstPath/{first}/{second}/{third}`, and the second parameter is wrong.

It will be replaced by `/myFirstPath/{first}/(.*)/{third}` and then by `/myFirstPath/.*`.

I think you need `"/\\{.*?}"` (non-greedy).

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+        return messages;
+    }
+
+    private JsonNode getRequestBody(HttpServletRequest request) {
+        try {
+            InputStream inputStream = request.getInputStream();
+            return mapper.readTree(inputStream);
+        } catch (IOException ex) {
+            if (!"GET".equals(request.getMethod()) && !"HEAD".equals(request.getMethod())) {
```
jpraet, line 210: I guess you get IOException when there is no body? See my earlier comment about only parsing the request body when necessary.

jpraet (changes requested):

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,217 @@
...
+        if (isNonExistingPath(ex)) {
+            return ProblemMediaType.INSTANCE.toResponse(new ResourceNotFoundProblem());
+        }
+        List<InputValidationIssue> issues = constructIssues(ex, request);
+        return ProblemMediaType.INSTANCE.toResponse(new BadRequestProblem(issues));
+    }
+
+    private List<InputValidationIssue> constructIssues(InvalidRequestException ex, HttpServletRequest request) {
+        Set<InputValidationIssue> issues = new LinkedHashSet<>();
+        JsonNode requestBody = getRequestBody(request);
```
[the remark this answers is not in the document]
smals-mavh: The annoying thing is that the inputstream in the HttpRequest can only be read once. And you'll need the reference multiple times when there is are multiple issues in the body.
Solved it with ArtomicReference. Now it's only set when it's actually needed.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
smals-mavh: Or when there is an actual error, I kept the error log, but removed the if block to verify the request method.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
smals-mavh: Good catch, also implemented testcases for this.

belgif-rest-problem-spring/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * This class is only active when Spring-Boot applications implemented the Atlassian swagger-request-validator.
+ * Maps the InvalidRequestException thrown by the swagger-request-validator to belgif problems.
+ */
+@RestControllerAdvice
+@ConditionalOnWebApplication
+@ConditionalOnClass(InvalidRequestException.class)
+@Order(1)
+// @Order(1) to take precedence over io.github.belgif.rest.problem.spring.ProblemExceptionHandler
+public class InvalidRequestExceptionHandler {
```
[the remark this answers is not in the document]
smals-mavh: Added testclasses with unit tests.
I couldn't easily mock the static methods in the utils class because adding the mockito-inline dependency breaks the other mocks.
I think this will work as well.

pvdbosch (approved):

jpraet (approved):

PR #120, opened 2024-10-19
Improve documentation

### LINKED ISSUE #117: [doc improvement] Create dedicated "how to use" section
I found that our users sometimes have a bit of trouble finding out from the documentation how to use the libraries, while they actually work almost entirely automagically.

I'd suggest to create a dedicated "how to use" section in the doc, because its sometimes now mixed with the "how it works" explanation.

It could have:
* maven coordinates for dependency to be added per platform
* Mention that it works out of the box and framework (JAX-RS, Spring RestController) exceptions are converted to appropriate Problem responses.
* Code samples of 5.1 and 5.2 (+ slight variants for Spring)

Caveats for Spring:
* client-side: autowired RestTemplate/WebClient/RestClient work out of the box, for non-autowired instances, use one of the appropriate "xxxCustomizer" classes
* Spring's built-in Problem support shouldn't be activated (remove `spring.mvc.problemdetails.enabled: true` config if present)

Then sections can follow for Problem exception classes (3.1), custom problems  (3.2) and openapi-generator.

Details of ProblemRegistry and which platform exceptions are converted (mappers, handlers, ...) don't generally need to be known by users and could be in later sections.

jpraet:
Fixes #117.

* Extract release notes to separate document (only include most recent releases in main doc)
* Improve new user experience:
  * Add dependency coordinates for Jakarta EE and Spring Boot
  * Add "Getting started" section for Jakarta EE and Spring Boot
  * Extract the internal implementation details to another section at the end


jflabatBCSS (approved):
This looks good for me. 

I'm wondering if we need to add examples too for each client library supported: microprofile rest, etc. But this is maybe more general than specific to belgif because I'm not familiar with all this libraries.

pvdbosch (approved):
Very good! I'll be stealing the block switches idea for the REST guide :)

PR #127, opened 2024-10-22
Add WWW-Authenticate HTTP response header to token-related problem types

### LINKED ISSUE #126: Support WWW-Authenticate HTTP response header
See https://github.com/belgif/rest-guide/issues/204.

Support for WWW-Authenticate HTTP response header should be added to:

* [NoAccessTokenProblem](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/NoAccessTokenProblem.java)
* [InvalidAccessTokenProblem](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java)
* [ExpiredAccessTokenProblem](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/ExpiredAccessTokenProblem.java)
* [MissingScopeProblem](https://github.com/belgif/rest-problem-java/blob/main/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java)

jpraet:
Fixes #126

### DIFF
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
@@ -1,8 +1,11 @@
 package io.github.belgif.rest.problem;
 
 import java.net.URI;
+import java.util.Collections;
+import java.util.Map;
 
 import io.github.belgif.rest.problem.api.ClientProblem;
+import io.github.belgif.rest.problem.api.HttpResponseHeaders;
 import io.github.belgif.rest.problem.api.ProblemType;
 
 /**
@@ -12,7 +15,7 @@ import io.github.belgif.rest.problem.api.ProblemType;
  *      https://www.belgif.be/specification/rest/api-guide/#invalid-access-token</a>
  */
 @ProblemType(InvalidAccessTokenProblem.TYPE)
-public class InvalidAccessTokenProblem extends ClientProblem {
+public class InvalidAccessTokenProblem extends ClientProblem implements HttpResponseHeaders {
 
     /**
      * The problem type.
@@ -47,9 +50,23 @@ public class InvalidAccessTokenProblem extends ClientProblem {
 
     private static final long serialVersionUID = 1L;
 
+    private String reason = "The access token is invalid";
+
     public InvalidAccessTokenProblem() {
         super(TYPE_URI, HREF, TITLE, STATUS);
         setDetail(DETAIL);
+    }
+
+    public InvalidAccessTokenProblem(String reason) {
+        this();
+        setDetail(DETAIL + ": " + reason);
+        this.reason = reason;
+    }
+
+    @Override
+    public Map<String, Object> getHttpResponseHeaders() {
+        return Collections.singletonMap(WWW_AUTHENTICATE,
+                String.format("Bearer error=\"invalid_token\", error_description=\"%s\"", reason));
     }
 
 }
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
@@ -51,16 +51,19 @@ public class InvalidAccessTokenProblem extends ClientProblem implements HttpResp
 
     private static final long serialVersionUID = 1L;
 
-    private String reason = "The access token is invalid";
+    private final String reason;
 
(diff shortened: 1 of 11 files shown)


belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
```diff
@@ -47,9 +51,43 @@ public class InvalidAccessTokenProblem extends ClientProblem {
 
     private static final long serialVersionUID = 1L;
 
+    private String reason = "The access token is invalid";
```
clone1612, line 54: Field can be made final

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java
```diff
@@ -99,4 +101,9 @@ public int hashCode() {
         return Objects.hash(super.hashCode(), requiredScopes);
     }
 
+    @Override
+    public Map<String, Object> getHttpResponseHeaders() {
+        return Collections.singletonMap(WWW_AUTHENTICATE, "Bearer error=\"insufficient_scope\"");
```
clone1612, line 106: scope(s) to indicate you might be missing multiple?

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/ExpiredAccessTokenProblemTest.java
```diff
@@ -18,6 +19,8 @@ void construct() {
         assertThat(problem.getStatus()).isEqualTo(401);
         assertThat(problem.getDetail())
                 .isEqualTo("The Bearer access token found in the Authorization HTTP header has expired");
+        assertThat(problem.getHttpResponseHeaders()).containsEntry(HttpResponseHeaders.WWW_AUTHENTICATE,
```
clone1612, line 22: Shouldn't we also validate the header is properly set in our integration tests?

clone1612 (changes requested):

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
jpraet: Only if we make some changes to the initialization / constructor chaining logic.

How do you see this? The default constructor calling `this("The access token is invalid");`?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java
jpraet: The error code "insufficient_scope" is part of the spec.
https://datatracker.ietf.org/doc/html/rfc6750#section-3.1

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/ExpiredAccessTokenProblemTest.java
jpraet: I'm not sure if this is really needed, as we already have an integration test for the Retry-After HTTP header, which uses the same .getHttpResponseHeaders() mechanism?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/InvalidAccessTokenProblem.java
clone1612: I was thinking about:

```
    private final String reason;

    public InvalidAccessTokenProblem() {
        this(DETAIL, "The access token is invalid");
    }

    public InvalidAccessTokenProblem(String reason) {
        this(DETAIL + ": " + reason, reason);
    }

    private InvalidAccessTokenProblem(String detail, String reason) {
        super(TYPE_URI, HREF, TITLE, STATUS);
        setDetail(detail);
        this.reason = reason;
    }
```

belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/ExpiredAccessTokenProblemTest.java
clone1612: Maybe not, I think it's personal preference. I usually treat the application as a "black box" testing all scenarios (not considering shared logic) but as you say it's tested through Retry-After.

clone1612 (approved):

PR #132, opened 2024-10-29
Add fluent problem properties

jpraet:


belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationProblemTest.java
```diff
@@ -109,6 +109,13 @@ void addInvalidParam() {
         assertThat(problem.getInvalidParams()).isUnmodifiable().containsExactly(invalidParam);
     }
 
+    @Test
+    void fluentProperties() {
+        InputValidationProblem problem = new MyInputValidationProblem()
+                .detail("detail message")
```
jpraet, line 115: If we don't override the fluent setters in InputValidationProblem, this chaining wouldn't work.

You would need to reverse the order to

```java
        InputValidationProblem problem = new MyInputValidationProblem()
                .issues(InputValidationIssues.requiredInput(InEnum.QUERY, "test"))
                .detail("detail message");
```

The self-referencing generics trick we used in https://github.com/belgif/rest-problem-java/pull/101/files cannot be applied here, because an exception class cannot be generic.

clone1612: I've played around with the implementation but wasn't able to find an alternative approach. Even looked at the [manifold](https://github.com/manifold-systems/manifold) compiler plugin which Pieter often mentions but even its crazy change features don't seem to often anything we can use here.

Simply a limitation of the language.

jpraet: I tried out another approach with interfaces and default methods to reduce the boilerplate.
One drawback is that an explicit `implements Fluent*Problem<SELF>` is required on all concrete problem classes that wish to have these fluent properties. Whereas in the previous approach the fluent properties would automatically be inherited from the super class (but with the "wrong" return type, requiring to override to allow proper method chaining).

What do you think @belgif/rest-problem-java?

PR #141, opened 2024-11-19
Add in(Input<?>) in InputValidationIssue for a fluent single input setter

### LINKED ISSUE #140: Add in(Input<?>) in InputValidationIssue for fluent single input setter
Some applications are referencing parameters using the `Input` class. There is currently no operation that allows to set a single input in a fluent manner.

We should propose a new operation in `InputValidationIssue` that works in the same manner as the `.in(in, name, value)` operation.

jflabatBCSS:
Fixes #140

### DIFF
--- a/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssueTest.java
+++ b/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssueTest.java
@@ -107,14 +107,10 @@ class InputValidationIssueTest {
         assertThat(issue.getName()).isNull();
         assertThat(issue.getValue()).isNull();
 
-        issue.in(Input.query("name", "value"));
+        issue = issue.in(Input.query("name", "value"));
         assertThat(issue.getIn()).isEqualTo(InEnum.QUERY);
         assertThat(issue.getName()).isEqualTo("name");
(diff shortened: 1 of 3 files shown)


belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/InputValidationIssueTest.java
```diff
@@ -90,14 +90,33 @@ void in() {
...
+        assertThat(issue.getIn()).isNull();
+        assertThat(issue.getName()).isNull();
+        assertThat(issue.getValue()).isNull();
+
+        issue.in(Input.query("name", "value"));
+        assertThat(issue.getIn()).isEqualTo(InEnum.QUERY);
+        assertThat(issue.getName()).isEqualTo("name");
+        assertThat(issue.getValue()).isEqualTo("value");
+
+        assertThat(new InputValidationIssue().in(Input.query("name", "value")).getIn()).isEqualTo(InEnum.QUERY);
```
jpraet, line 115: Aren't these 3 lines redundant with the 3 lines above?

jpraet (changes requested):
Can you add an entry to the release notes?

For version 0.11

jpraet (approved):

PR #145, opened 2024-12-05
Standardize urn:problem-type:belgif:input-validation:referencedResourceNotFound

jpraet:
https://github.com/belgif/rest-guide/issues/126.


jpraet:
@pvdbosch wasn't there also some consensus in the workgroup around `urn:problem-type:belgif:input-validation:invalidStructure`?

pvdbosch (approved):

pvdbosch:
> @pvdbosch wasn't there also some consensus in the workgroup around `urn:problem-type:belgif:input-validation:invalidStructure`?

There was consensus that standardization is needed for this use case, but there are still some points to discuss:
*  the fine line between 'schemaViolation' and 'invalidStructure': is such a distinction useful for an API user; does the distinction depend too much on implementation of validation?
* check if everyone is OK with the name of the issue type

PR #146, opened 2024-12-08
Add support for Quarkus

jpraet:


belgif-rest-problem-it/belgif-rest-problem-quarkus-it/src/main/java/io/github/belgif/rest/problem/quarkus/it/FrontendImpl.java
```diff
@@ -0,0 +1,232 @@
...
+
+    private Backend microprofileClient;
+
+    private jakarta.ws.rs.client.Client jaxRsClient;
+
+    @PostConstruct
+    public void initialize() {
+        this.baseUri = URI.create("http://" + serverRequest.localAddress().host() + ":"
+                + serverRequest.localAddress().port() + "/quarkus");
+        this.microprofileClient = RestClientBuilder.newBuilder()
```
clone1612, line 48: Quarkus [REST client guide](https://quarkus.io/guides/rest-client) describes also these 2 ways of creating clients:
* `@RegisterRestClient` annotated interface injected in resource class with `@RestClient`
* Using the `QuarkusRestClientBuilder` 

While underlying microprofile is used verifying that constructed through these different method behavior is still the same adds a bit of certainty (properly configured for all methods described in the Quarkus guides).

clone1612 (approved):

PR #148, opened 2024-12-10
Improve WWW-Authenticate headers

jpraet:
See https://github.com/belgif/rest-guide/pull/215 for more context.

* Support setting the "realm" attribute
* Add "error_description" and "scope" attributes for missing_scope

### DIFF
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java
@@ -53,6 +53,8 @@ public class MissingScopeProblem extends ClientProblem implements HttpResponseHe
 
     private final List<String> requiredScopes = new ArrayList<>();
 
+    private String realm;
+
     public MissingScopeProblem() {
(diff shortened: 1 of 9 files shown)


pvdbosch (approved):

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/MissingScopeProblem.java
```diff
@@ -62,6 +64,11 @@ public MissingScopeProblem(String... requiredScopes) {
         setRequiredScopes(requiredScopes);
     }
```
jflabatBCSS, line 66: You sometimes provide a constructor with the realm parameter ? Why don't you propose it here ?

jpraet: The realm is a string, so it will conflict with `String... requiredScopes`

jflabatBCSS: Ok, I didn't see that constructor in the track changes

PR #174, opened 2025-02-03
Add support for Quarkus in native mode

jpraet:


jflabatBCSS (approved):
From a Java perspective, I don't see any issue in my review

PR #177, opened 2025-02-05
Fix array handling in in JacksonUtil

### LINKED ISSUE #176: Bug: Arrays are not handled correctly in JacksonUtil.toBadRequestProblem
In JacksonUtil, the path from the MismatchInputException is used to build the name property in the schemaViolation issue. But when the path contains an array, the fieldName is null which lead to a path like "array.null.subpath".

jflabatBCSS:
Fixes #176

### DIFF
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/internal/JacksonUtil.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/internal/JacksonUtil.java
@@ -26,7 +26,7 @@ public class JacksonUtil {
     public static BadRequestProblem toBadRequestProblem(JsonMappingException e) {
         StringBuilder name = new StringBuilder();
         for (Reference reference : e.getPath()) {
-            if (reference.getFrom() != null && reference.getFrom() instanceof List) {
+            if (reference.getFrom() instanceof List) {
                 name.append("[").append(reference.getIndex()).append("]");
             } else {
                 if (name.length() > 0) {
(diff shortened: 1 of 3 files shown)


belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/internal/JacksonUtil.java
```diff
@@ -24,9 +24,19 @@ private JacksonUtil() {
      * @return the BadRequestProblem
      */
     public static BadRequestProblem toBadRequestProblem(JsonMappingException e) {
-        String name = e.getPath().stream().map(Reference::getFieldName).collect(Collectors.joining("."));
+        StringBuilder name = new StringBuilder();
+        for (Reference reference : e.getPath()) {
+            if (reference.getFrom() != null && reference.getFrom() instanceof List) {
```
jpraet, line 29: Sonar [suggests to remove the null check](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=177&id=belgif_rest-problem-java&open=AZTVTJO5vLz2aA1LC8Ep).

jpraet (changes requested):
Could you add an entry to the [release notes](https://github.com/belgif/rest-problem-java/blob/main/src/main/asciidoc/release-notes.adoc)? (0.15)

I suggest something like "Fix input path for MismatchedInputException on array properties"

jpraet (approved):

PR #236, opened 2025-09-24
Add href property with documentation URL for issue types

jpraet:


belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/it/AbstractRestProblemIT.java
```diff
@@ -170,6 +170,8 @@ public void constraintViolationMissingRequiredQueryParameter() {
                 .body("type", equalTo("urn:problem-type:belgif:badRequest"))
                 .body("detail", equalTo("The input message is incorrect"))
                 .body("issues[0].type", equalTo("urn:problem-type:belgif:input-validation:schemaViolation"))
+                .body("issues[0].href",
```
clone1612, line 173: Should we also add the check to the `requestValidator()` test (invalidInput)?

clone1612 (approved):

PR #258, opened 2025-11-24
Correctly determine source for @BeanParam validation issues

### LINKED ISSUE #253: Java ee: POJO annotated with @BeanParam reports as BODY
Problem

When a validation fails on a POJO annotated with @BeanParam, it reports the attribute as in the BODY, rather than in the QUERY.

```
List<Person> getPersons(
                   @QueryParam("valid") @NotNull Boolean valid, // ok - reports as query problem
                   @Valid @BeanParam PersonCriteria personCriteria);

...
public class PersonCriteria {

    @NotBlank
    @QueryParam("lastName") // nok - should report as query problem, not body
    private String lastName;

...
```

jpraet:
Fixes #253

### DIFF
--- a/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/internal/ConstraintViolationUtil.java
+++ b/belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/internal/ConstraintViolationUtil.java
@@ -79,10 +79,7 @@ public class ConstraintViolationUtil {
 
     private static Input<Object> determineInput(ConstraintViolation<?> violation,
             MethodNode methodNode, List<Node> propertyPath, List<String> propertyName) {
-        Input<Object> input = new Input<>();
-        input.setIn(InEnum.BODY);
-        input.setName(String.join(".", propertyName));
-        input.setValue(violation.getInvalidValue());
+        Input<Object> input = Input.body(String.join(".", propertyName), violation.getInvalidValue());
         Node last = propertyPath.get(propertyPath.size() - 1);
         Node parent = propertyPath.size() > 1 ? propertyPath.get(propertyPath.size() - 2) : null;
         if (last.getKind() == ElementKind.PARAMETER) {


belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/internal/ConstraintViolationUtil.java
```diff
@@ -45,56 +48,81 @@ private ConstraintViolationUtil() {
...
+            propertyPath.add(p);
             if (p.getKind() == ElementKind.METHOD) {
                 methodNode = p.as(MethodNode.class);
             }
             if ((p.getKind() != ElementKind.METHOD && p.getKind() != ElementKind.PARAMETER)
                     || !pathIterator.hasNext()) {
                 // skip leading METHOD or PARAMETER path segments,
                 // e.g. "sector" instead of "createPartner.partner.sector"
-                propertyPath.add(p);
+                propertyName.add(p.toString());
```
jhemelhof: I'm not sure if this is correct for the BeanParam scenario. The `JaxRsParameterNameProvider` only uses the annotation value for constructor and method parameters.

jpraet: I made the changes to retrieve the input name from the JAX-RS param annotations.

jpraet: We should probably add FormParam, MatrixParam and CookieParam to the SOURCE_MAPPING.

* FormParam can map to "in": "body", see https://spec.openapis.org/oas/v3.0#example-basic-multipart-form
* MatrixParam can map to "in": "path", see https://spec.openapis.org/oas/v3.0#style-values
* CookieParam should map to "in": "cookie", see https://spec.openapis.org/oas/v3.0#parameter-locations
  -> but then it should be added to the enum in problem-v1.yaml: https://github.com/belgif/openapi-problem/blob/af2d348540a7e4aa58f19a100dcd1b89fd303b74/src/main/openapi/problem/v1/problem-v1.yaml#L141
What do we do in the meantime? Leave "in" empty?

PR #268, opened 2025-12-16
New exception mappers for JSON parse and ParamConverter exceptions

### LINKED ISSUE #267: New exception mappers for JSON parse and ParamConverter exceptions
io.github.belgif.rest.problem.ee.jaxrs.JsonMappingExceptionMapper

Maps JsonMapping exceptions to issue detail messages

    private String getReason(Class<? extends JsonMappingException> clazz) {
        switch (clazz.getSimpleName()) {
            case "InvalidFormatException":
                return "value has invalid format";
            case "InvalidNullException":
                return "property should not be null";
            case "InvalidTypeIdException":
                return "invalid discriminator property";
            case "IgnoredPropertyException":
                return "unexpected property";
            case "UnrecognizedPropertyException":
                return "unrecognized property";
            case "ValueInstantiationException":
                return "invalid value for property";
            default:
                return "invalid property value";
        }
    }

io.github.belgif.rest.problem.ee.jaxrs.JsonParseExceptionMapper

Maps JsonParsing exceptions to issue detail messages

    private String getReason(Class<? extends JsonParseException> clazz) {
        if (clazz == JsonEOFException.class) {
            return "invalid json data (end-of-input reached unexpectedly)";
        } else {
            return "invalid json data";
        }
    }

io.github.belgif.rest.problem.ee.jaxrs.NotFoundExceptionMapper

{@link javax.ws.rs.ext.ParamConverter} errors are wrapped into a {@link javax.ws.rs.NotFoundException} in JBoss EAP. This ExceptionMapper unwraps them into 400 Bad Request instead of 404.

wsalembi:
Fixes #267.

### DIFF
--- a/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/ee/jaxrs/BadRequestExceptionMapperTest.java
+++ b/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/ee/jaxrs/BadRequestExceptionMapperTest.java
@@ -1,7 +1,6 @@
 package io.github.belgif.rest.problem.ee.jaxrs;
 
 import static org.assertj.core.api.Assertions.*;
-import static org.junit.jupiter.api.Assertions.*;
 
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.core.Response;
@@ -32,8 +31,8 @@ class BadRequestExceptionMapperTest {
         assertThat(response.getEntity()).isInstanceOf(BadRequestProblem.class);
         BadRequestProblem problem = (BadRequestProblem) response.getEntity();
         InputValidationIssue issue = problem.getIssues().get(0);
-        assertEquals(InEnum.HEADER, issue.getIn());
-        assertEquals("startDate_gt", issue.getName());
+        assertThat(issue.getIn()).isEqualTo(InEnum.HEADER);
+        assertThat(issue.getName()).isEqualTo("startDate_gt");
     }
 
     @Test
@@ -47,8 +46,8 @@ class BadRequestExceptionMapperTest {
         assertThat(response.getEntity()).isInstanceOf(BadRequestProblem.class);
         BadRequestProblem problem = (BadRequestProblem) response.getEntity();
         InputValidationIssue issue = problem.getIssues().get(0);
-        assertEquals(InEnum.HEADER, issue.getIn());
-        assertEquals("startDate_gt", issue.getName());
+        assertThat(issue.getIn()).isEqualTo(InEnum.HEADER);
+        assertThat(issue.getName()).isEqualTo("startDate_gt");
     }
 
     @Test
--- a/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/ee/jaxrs/NotFoundExceptionMapperTest.java
+++ b/belgif-rest-problem-java-ee/src/test/java/io/github/belgif/rest/problem/ee/jaxrs/NotFoundExceptionMapperTest.java
@@ -1,7 +1,6 @@
 package io.github.belgif.rest.problem.ee.jaxrs;
 
 import static org.assertj.core.api.Assertions.*;
-import static org.junit.jupiter.api.Assertions.*;
 
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.core.Response;
@@ -33,8 +32,8 @@ class NotFoundExceptionMapperTest {
         assertThat(response.getEntity()).isInstanceOf(BadRequestProblem.class);
         BadRequestProblem problem = (BadRequestProblem) response.getEntity();
         InputValidationIssue issue = problem.getIssues().get(0);
-        assertEquals(InEnum.QUERY, issue.getIn());
-        assertEquals("startDate_gt", issue.getName());
+        assertThat(issue.getIn()).isEqualTo(InEnum.QUERY);
+        assertThat(issue.getName()).isEqualTo("startDate_gt");
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/jaxrs/JsonParseExceptionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+@Provider
+public class JsonParseExceptionMapper implements ExceptionMapper<JsonParseException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JsonParseExceptionMapper.class);
+
+    @Override
+    public Response toResponse(JsonParseException exception) {
+        LOGGER.warn(exception.getMessage(), exception);
+        try {
+            ProblemConfig.setExtIssueTypesEnabled(true);
```
jpraet: We shouldn't alter the setExtIssueTypesEnabled flag here, it is up to the applicaiton owner to enable/disable.

Based on that flag, issue type "urn:problem-type:belgif-ext:input-validation:invalidStructure" (enabled) or "urn:problem-type:belgif:input-validation:invalidInput" (disabled) will be used.

belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/jaxrs/NotFoundExceptionMapper.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(NotFoundExceptionMapper.class);
+
+    private static final Pattern PARAM_PATTERN =
+            Pattern.compile("(javax|jakarta).ws.rs.(Path|Query|Header|Bean|Form|Cookie|Matrix)Param\\(\"(.*)\"\\)");
+
+    @Override
+    public Response toResponse(NotFoundException exception) {
+        LOGGER.warn(exception.getMessage(), exception);
+
+        if (exception.getCause() instanceof BadRequestProblem) {
```
jpraet: I'm a bit confused about this. BadRequestProblem gets wrapped in javax.ws.rs.NotFoundException?
Would be good to have integration tests to demonstrate the use cases and desired behavior, also for the other 2 exception mappers.

There is a regression in one of the integration tests: it used to result in 404 not found but now gets 500 internal error.

belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/it/AbstractRestProblemEEIT.java
```diff
@@ -40,4 +40,89 @@ public void constraintViolationBeanParam() {
...
+                        equalTo("https://www.belgif.be/specification/rest/api-guide/issues/schemaViolation.html"))
+                .body("issues[0].title", equalTo("Input value is invalid with respect to the schema"))
+                .body("issues[0].detail", equalTo("date has invalid format"))
+                .body("issues[0].in", equalTo("query"))
+                .body("issues[0].name", equalTo("date"))
+                .body("issues[0].value", equalTo("2025-13-12"));
+    }
+
+    @Test
+    public void invalidJsonEOF() {
```
jpraet: On Spring Boot, the "invalidJson", "invalidJsonEOF" and "invalidJsonFormat" tests result in 

```json
{
    "title": "Bad Request",
    "status": 400,
    "issues": [
        {
            "href": "https://www.belgif.be/specification/rest/api-guide/issues/schemaViolation.html",
            "detail": "JSON parse error",
            "in": "body"
        }
    ]
}
```

Would be best to align the behavior. Which make most sense? "urn:problem-type:belgif:input-validation:schemaViolation" or "urn:problem-type:belgif:input-validation:invalidInput"?

belgif-rest-problem-java-ee/src/main/java/io/github/belgif/rest/problem/ee/jaxrs/NotFoundExceptionMapper.java
jpraet: Indeed, from JAX-RS spec: https://jakarta.ee/specifications/restful-ws/4.0/jakarta-restful-ws-spec-4.0#resource_field

> A WebApplicationException thrown during construction of field or property values using any of the 5 steps listed above is processed directly as described in [Exceptions](https://jakarta.ee/specifications/restful-ws/4.0/jakarta-restful-ws-spec-4.0#method_exc). Other exceptions thrown during construction of field or property values using any of the 5 steps listed above are treated as client errors: if the field or property is annotated with @MatrixParam, @QueryParam or @PathParam then an implementation MUST generate an instance of NotFoundException (404 status) that wraps the thrown exception and no entity; if the field or property is annotated with @HeaderParam or @CookieParam then an implementation MUST generate an instance of BadRequestException (400 status) that wraps the thrown exception and no entity. Exceptions MUST be processed as described in [Exceptions](https://jakarta.ee/specifications/restful-ws/4.0/jakarta-restful-ws-spec-4.0#method_exc).

So unwrapping BadRequestProblem cause of javax.ws.rs.NotFoundException indeed makes sense. The enrichment of the "in" and "name" properties of the input validation issue code does seem a bit fragile though (parsing RESTEASY specific exception message like `RESTEASY003870: Unable to extract parameter from http request: jakarta.ws.rs.QueryParam("startDate_gt") value is '2006-087-01'`). This may not work on other JAX-RS runtimes.
I am wondering if it isn't possible to directly fill in the correct "in" and "name" inside the ParamConverter?
[ParamConverterProvider](https://jakarta.ee/specifications/platform/11/apidocs/jakarta/ws/rs/ext/paramconverterprovider) has access to the annotations, so could pass those on to the ParamConverter constructor.

Also interesting to add a test with @HeaderParam or @CookieParam. Possibly the current approach won't work yet for those, as the spec says those result in javax.ws.rs.BadRequestException instead of javax.ws.rs.NotFoundException.

belgif-rest-problem-it/belgif-rest-problem-it-common/src/main/java/io/github/belgif/rest/problem/it/AbstractRestProblemEEIT.java
jpraet: @pvdbosch which issue type do you think is most appropriate for a malformed JSON request body?

* urn:problem-type:belgif:input-validation:invalidInput (and urn:problem-type:belgif-ext:input-validation:invalidStructure when belgif-ext is enabled)
* urn:problem-type:belgif:input-validation:schemaViolation

PR #272, opened 2025-12-22
Add spring boot 4 support

smals-mavh:
Added Spring Boot 4 support

### DIFF
--- a/belgif-rest-problem-spring-boot-common/src/test/java/io/github/belgif/rest/problem/spring/AbstractInvalidRequestExceptionHandlerTest.java
+++ b/belgif-rest-problem-spring-boot-common/src/test/java/io/github/belgif/rest/problem/spring/AbstractInvalidRequestExceptionHandlerTest.java
@@ -219,7 +219,7 @@ class AbstractInvalidRequestExceptionHandlerTest {
         Problem problem = handler.handleInvalidRequestException(exception, request).getBody();
         assertThat(problem).isInstanceOf(BadRequestProblem.class);
         BadRequestProblem badRequestProblem = (BadRequestProblem) problem;
-        assertThat(badRequestProblem.getIssues().size()).isEqualTo(2);
+        assertThat(badRequestProblem.getIssues()).hasSize(2);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


belgif-rest-problem-spring-boot-4/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,79 @@
...
+        JsonNode requestBody = getRequestBody(requestBodyReference, request, mapper);
+        if (requestBody == null) {
+            return null;
+        }
+        JsonNode valueNode = requestBody.at(JsonPointer.compile(name));
+        return extractString(valueNode, mapper);
+    }
+
+    private static JsonNode getRequestBody(AtomicReference<JsonNode> requestBodyReference, HttpServletRequest request,
+            ObjectMapper mapper) {
```
jhemelhof: Does this method need to be static? Doesn't seem to be the case, and that way you can use the class mapper instead of passing it as a parameter.

belgif-rest-problem-spring-boot-4/src/main/java/io/github/belgif/rest/problem/spring/InvalidRequestExceptionHandler.java
```diff
@@ -0,0 +1,79 @@
...
+        if (requestBodyReference.get() == null) {
+            try {
+                InputStream inputStream = request.getInputStream();
+                requestBodyReference.set(mapper.readTree(inputStream));
+            } catch (IOException ex) {
+                LOGGER.error("Error reading input stream", ex);
+            } catch (NullPointerException ex) {
```
jhemelhof: Do we want to catch a NPE? Wouldn't it be cleaner to do that in a separate if-check?

belgif-rest-problem-spring-boot-common/src/test/java/io/github/belgif/rest/problem/spring/AbstractInvalidRequestExceptionHandlerTest.java
```diff
@@ -0,0 +1,274 @@
...
+        ValidationReport.Message message = ValidationReport.Message.create("my.validation.issue", "test message")
+                .build().withNestedMessages(messages);
+
+        InvalidRequestException exception = new InvalidRequestException(ValidationReport.from(message));
+        HttpServletRequest request = new MockHttpServletRequest();
+
+        Problem problem = handler.handleInvalidRequestException(exception, request).getBody();
+        assertThat(problem).isInstanceOf(BadRequestProblem.class);
+        BadRequestProblem badRequestProblem = (BadRequestProblem) problem;
+        assertThat(badRequestProblem.getIssues().size()).isEqualTo(1);
```
jhemelhof: .hasSize(1)

jhemelhof (changes requested):

PR #274, opened 2025-12-22
Don't rely on schema validation

jpraet:
belgif-rest-problem-validator currently expects some errors to be caught by openapi schema validation, but that's not always enabled. So these validations need to be duplicated in the library.

### DIFF
--- a/belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/EnterpriseNumberValidatorTest.java
+++ b/belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/EnterpriseNumberValidatorTest.java
@@ -18,7 +18,7 @@ class EnterpriseNumberValidatorTest {
     }
 
     @ParameterizedTest
-    @ValueSource(strings = { "test", "54321", "2111111111", "0884303370" })
+    @ValueSource(strings = { "test", "54321", "20000000032", "2111111111", "0884303370" })
     void nok(String value) {
         assertThat(new EnterpriseNumberValidator(Input.body("test", value)).validate())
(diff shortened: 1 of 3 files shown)


belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/EnterpriseNumberValidatorTest.java
```diff
@@ -15,10 +17,11 @@ void ok() {
         assertThat(new EnterpriseNumberValidator(Input.body("test", "0884303369")).validate()).isEmpty();
     }
 
-    @Test
-    void nok() {
-        assertThat(new EnterpriseNumberValidator(Input.body("test", "2111111111")).validate())
-                .contains(InputValidationIssues.invalidEnterpriseNumber(BODY, "test", "2111111111"));
+    @ParameterizedTest
+    @ValueSource(strings = { "test", "54321", "2111111111", "0884303370" })
```
jflabatBCSS: Maybe add a 10 length string not starting with 0 or 1 and with a valid checksum

belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/EstablishmentUnitNumberValidatorTest.java
```diff
@@ -15,10 +17,11 @@ void ok() {
         assertThat(new EstablishmentUnitNumberValidator(Input.body("test", "2297964444")).validate()).isEmpty();
     }
 
-    @Test
-    void nok() {
```
jflabatBCSS: Maybe add a value starting with 0 or 1 and with a valid checksum

jflabatBCSS (approved):

PR #282, opened 2026-01-09
Disable stacktrace on Problem exceptions

### LINKED ISSUE #278: Disable stacktrace on Problem exceptions
Java exceptions are expensive. Especially [constructing](https://www.baeldung.com/java-exceptions-performance#3-creating-an-exception-without-throwing-it) them and initializing their stacktrace.

The rest-problem-java library treats problems as exceptions (Problem extends RuntimeException). Typically, the problems are thrown only to have them then converted to a proper "application/problem+json" response. The stacktrace is not really necessary.

This [overloaded Throwable constructor](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#Throwable-java.lang.String-java.lang.Throwable-boolean-boolean-) accepts a writableStackTrace boolean flag:

> If the writable stack trace is false, this constructor will not call [fillInStackTrace()](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#fillInStackTrace--), a null will be written to the stackTrace field, and subsequent calls to fillInStackTrace and [setStackTrace(StackTraceElement[])](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#setStackTrace-java.lang.StackTraceElement:A-) will not set the stack trace. If the writable stack trace is false, [getStackTrace()](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#getStackTrace--) will return a zero length array.

So we could set this flag to false for a performance gain.

If necessary, we could add a `io.github.belgif.rest.problem.stacktrace-disabled` flag (default true) which allows users to opt-out if they do want their problem exceptions to have a stacktrace.

jpraet:
Fixes #278

### DIFF
--- a/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/ProblemTest.java
+++ b/belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/ProblemTest.java
@@ -139,4 +139,11 @@ class ProblemTest {
         }
     }
 
+    @Test
+    void stackTraceDisabledByDefault() {
+        ProblemConfig.reset();
+        Problem problem = new MyProblem();
+        assertThat(problem.getStackTrace()).isEmpty();
+    }
+
 }
--- a/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/config/ProblemConfig.java
+++ b/belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/config/ProblemConfig.java
@@ -19,7 +19,7 @@ public class ProblemConfig {
 
     private static final boolean DEFAULT_I18N_ENABLED = true;
 
-    private static final boolean DEFAULT_STACK_TRACE_ENABLED = true;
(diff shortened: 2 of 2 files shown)


belgif-rest-problem/src/test/java/io/github/belgif/rest/problem/api/ProblemTest.java
```diff
@@ -113,7 +115,28 @@ void getMessage() {
...
+    }
+
+    @Test
+    void stackTraceDisabled() {
+        try {
+            ProblemConfig.setStackTraceEnabled(false);
+            Problem problem = new MyProblem();
+            assertThat(problem.getStackTrace()).isEmpty();
+        } finally {
+            ProblemConfig.reset();
```
usku01: maybe we should add a single 'reset'  unit test to make sure that all default values are what we expect after the reset?

PR #348, opened 2026-06-19
Use JsonPointer syntax

### LINKED ISSUE #344: Use JSON pointer syntax in "name" field for request body input validation issues
See https://github.com/belgif/rest-guide/issues/290.

e.g. `"/boardMembers/0/ssin"` instead of `"boardMembers[0].ssin"`

We could provide a configuration flag `io.github.belgif.rest.problem.json-pointer-enabled` which defaults to `true` but can be switched to `false` to keep using the legacy JSON path flavor.

usku01:
Fixes #344


jpraet:
JsonPointer syntax only needs to be used for InEnum.BODY.
A query, path or header param with name "foo.bar" should be accepted.

usku01:
> JsonPointer syntax only needs to be used for InEnum.BODY. A query, path or header param with name "foo.bar" should be accepted.

What if we have a query parameter which would be a list, shouldn't we add the index in the name? And what I mean by that is of the query param is named 'parameter.id', if we have an issue with the second member of the list, then the name in the issue should be 'parameter.id/1'?

usku01:
> > JsonPointer syntax only needs to be used for InEnum.BODY. A query, path or header param with name "foo.bar" should be accepted.
> 
> What if we have a query parameter which would be a list, shouldn't we add the index in the name? And what I mean by that is of the query param is named 'parameter.id', if we have an issue with the second member of the list, then the name in the issue should be 'parameter.id/1'?

I have applied changes to the implementation to allow "." in name

jpraet:
> > JsonPointer syntax only needs to be used for InEnum.BODY. A query, path or header param with name "foo.bar" should be accepted.
> 
> What if we have a query parameter which would be a list, shouldn't we add the index in the name? And what I mean by that is of the query param is named 'parameter.id', if we have an issue with the second member of the list, then the name in the issue should be 'parameter.id/1'?

I think it's fine to keep using `parameter.id[1]` there.
JSON Pointer is specifically for pointing to a specific value within a JSON object.

belgif-rest-problem-it/belgif-rest-problem-jakarta-ee-it/src/main/java/io/github/belgif/rest/problem/LocalDateConverter.java
```diff
@@ -31,7 +32,9 @@ protected LocalDate fromString(InEnum in, String name, String value) {
         try {
             return LocalDate.parse(value, LOCAL_DATE_FORMATTER);
         } catch (DateTimeParseException e) {
-            throw new BadRequestProblem(schemaViolation(in, name, value, "date has invalid format"));
+            throw new BadRequestProblem(
+                    schemaViolation(in, JsonPointerUtil.transformName(in, name), value,
```
jpraet: Not sure the transformName is needed here (and in the other similar LocalDateConverter classes).
ParamConverter is mostly for query / path params. Will we ever encounter InEnum.BODY here?
There is a best-effort mapping from FormParam.class, but that is because `in: form` does not exist.
Anyway, that will still get auto-converted with a warning if needed.

belgif-rest-problem-java-ee-server/src/main/java/io/github/belgif/rest/problem/ee/server/internal/ConstraintViolationUtil.java
```diff
@@ -101,6 +109,14 @@ private static Input<Object> determineInput(ConstraintViolation<?> violation,
                 }
             }
         }
+
+        if (ProblemConfig.isJsonPointerEnabled() && input.getName() != null && input.getIn() != InEnum.BODY
```
jpraet: What if we kept this method as it was before, and only put this here:

```java
if (input.getIn() == InEnum.BODY && ProblemConfig.isJsonPointerEnabled()) {
  input.setName(JsonPointerUtil.transformName(input.getIn(), input.getName());
}
```

it's also not ideal as we would first create name in dot-notation and then replace by JSON Pointer,
but I don't really like having to remove the unwanted leading '/' either 🤔

belgif-rest-problem-java-ee-server/src/test/java/io/github/belgif/rest/problem/ee/server/jaxrs/BadRequestExceptionMapperTest.java
```diff
@@ -25,14 +25,14 @@ void badRequestException() {
     @Test
     void badRequestProblem() {
         BadRequestProblem cause = new BadRequestProblem(
-                InputValidationIssues.schemaViolation(InEnum.HEADER, "startDate_gt", "2006-087-01",
```
jpraet, line 28: I think this can be reverted? startDate_gt is a proposed Belgif standard.

(it's meant for query params though, not for HEADER, which is a bit weird)

belgif-rest-problem-java-ee-server/src/test/java/io/github/belgif/rest/problem/ee/server/jaxrs/NotFoundExceptionMapperTest.java
```diff
@@ -26,14 +26,14 @@ void notFoundException() {
     @Test
     void badRequestProblem() {
         BadRequestProblem cause = new BadRequestProblem(
-                InputValidationIssues.schemaViolation(InEnum.QUERY, "startDate_gt", "2006-087-01",
```
jpraet, line 29: I think this can be reverted?

belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/PeriodValidatorTest.java
```diff
@@ -15,35 +15,35 @@ class PeriodValidatorTest {
...
+                Input.body("/period", new InputPeriod(LocalDate.of(2023, 10, 12), LocalDate.of(2023, 10, 12))))
                         .validate()).isEmpty();
         // startDate < endDate
         assertThat(new PeriodValidator(
-                Input.body("period", new InputPeriod(LocalDate.of(2023, 10, 11), LocalDate.of(2023, 10, 12))))
+                Input.body("/period", new InputPeriod(LocalDate.of(2023, 10, 11), LocalDate.of(2023, 10, 12))))
                         .validate()).isEmpty();
         // no startDate and endDate
-        assertThat(new PeriodValidator(Input.body("criteria.periods.period", new InputPeriod(null, null)))
+        assertThat(new PeriodValidator(Input.body("/criteria/periods/period", new InputPeriod(null, null)))
```
jpraet: If we want this to be more realistic it should probably be /criteria/periods/1/period

belgif-rest-problem-validator/src/test/java/io/github/belgif/rest/problem/validation/RequestValidatorTest.java
```diff
@@ -314,22 +325,28 @@ void refDataPredicateInvalid() {
                 InputValidationIssues.referencedResourceNotFound(QUERY, "refData", "x"));
     }
 
-    @Test
-    void refDatasCollectionValid() {
+    @ParameterizedTest
```
jpraet: Aren't these ParameterizedTests only relevant for tests using body inputs?

belgif-rest-problem-java-ee-server/src/main/java/io/github/belgif/rest/problem/ee/server/internal/ConstraintViolationUtil.java
usku01: > What if we kept this method as it was before, and only put this here:
> 
> ```java
> if (input.getIn() == InEnum.BODY && ProblemConfig.isJsonPointerEnabled()) {
>   input.setName(JsonPointerUtil.transformName(input.getIn(), input.getName());
> }
> ```
> 
> it's also not ideal as we would first create name in dot-notation and then replace by JSON Pointer, but I don't really like having to remove the unwanted leading '/' either 🤔

I did not like either having to remove the '/', your proposition is better

belgif-rest-problem-it/belgif-rest-problem-jakarta-ee-it/src/main/java/io/github/belgif/rest/problem/LocalDateConverter.java
usku01: > Not sure the transformName is needed here (and in the other similar LocalDateConverter classes). ParamConverter is mostly for query / path params. Will we ever encounter InEnum.BODY here? There is a best-effort mapping from FormParam.class, but that is because `in: form` does not exist. Anyway, that will still get auto-converted with a warning if needed.

Ok, we can do it like that.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/internal/JsonPointerUtil.java
```diff
@@ -0,0 +1,76 @@
...
+import io.github.belgif.rest.problem.api.InEnum;
+import io.github.belgif.rest.problem.config.ProblemConfig;
+
+/**
+ * Internal utility class for JSON Pointer expressions.
+ */
+public class JsonPointerUtil {
+
+    // e.g: /, field, /field, /field/0, /field/0/nested
+    private static final Pattern JSON_POINTER_PATTERN = Pattern.compile("/+[a-zA-Z0-9-]*+(/[a-zA-Z0-9-]++)*+");
```
jpraet: Maybe we are too strict with this regex (e.g. allowing underscores).

Also, technically the empty string "" is also a valid JSON Pointer. It can be used to point to the body itself.

Currently empty string will be converted to null in transformName.

belgif-rest-problem-validator/src/main/java/io/github/belgif/rest/problem/validation/AbstractRequestValidator.java
```diff
@@ -116,7 +118,9 @@ public SELF ssins(Input<List<String>> ssins) {
         if (ssins != null && ssins.getValue() != null) {
             int index = 0;
             for (String ssin : ssins.getValue()) {
-                ssin(new Input<>(ssins.getIn(), ssins.getName() + "[" + index + "]", ssin));
+                String name = JsonPointerUtil.transformName(ssins.getIn(),
```
jpraet: Shouldn't we remove the transformName call here?

We want auto-conversion with a warning when this is called with `Input.body("ssins", myList);` instead of `"/ssins"` no? Same for the refData methods below.

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssues.java
```diff
@@ -139,8 +140,10 @@ public static InputValidationIssue referencedResourceNotFound(InEnum in, String
      * @param <T> The type of the reference
      */
     public static <T> InputValidationIssue referencedResourceNotFound(InEnum in, String name, T value, List<T> source) {
-        String nameWithIndex = name + "[" + source.indexOf(value) + "]";
-        return referencedResourceNotFound(in, nameWithIndex, value);
+        String indexFormat = ProblemConfig.isJsonPointerEnabled() && in == InEnum.BODY ? ("/" + source.indexOf(value))
+                : ("[" + source.indexOf(value) + "]");
+        String nameWithIndex = name + indexFormat;
+        return referencedResourceNotFound(in, JsonPointerUtil.transformName(in, nameWithIndex), value);
```
jpraet: Same here, shouldn't we remove the transformName call and rely on auto-conversion with warning instead?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/internal/JsonPointerUtil.java
jpraet: Maybe it's sufficient to check if it's either an empty string or starts with /.
We don't really need full JSON Pointer syntax validation, it's only a heuristic to check whether we should trigger auto-conversion.

belgif-rest-problem-java-ee-server/src/test/java/io/github/belgif/rest/problem/ee/server/internal/ConstraintViolationUtilTest.java
jflabatBCSS, line 60: You could have used ParametrizedTest to pass the jsonPointer enable flag. But like this is fine for me to. You can keep what you prefer.

belgif-rest-problem-validator/src/main/java/io/github/belgif/rest/problem/validation/AbstractRequestValidator.java
jflabatBCSS: Is transformName not already handling jsonPath indexes ?

belgif-rest-problem/src/main/java/io/github/belgif/rest/problem/api/InputValidationIssue.java
```diff
@@ -186,6 +197,25 @@ private boolean hasInNameValue() {
...
+    private void autoConvertToJsonPointerIfNeeded() {
+        if (ProblemConfig.isJsonPointerEnabled()) {
+            name = autoConvertToJsonPointerIfNeeded(in, name);
+            for (Input<?> input : inputs) {
+                input.setName(autoConvertToJsonPointerIfNeeded(input.getIn(), input.getName()));
+            }
+        }
+    }
+
+    private String autoConvertToJsonPointerIfNeeded(InEnum in, String propertyPath) {
```
jflabatBCSS: It is strange to overload a method but that do not really to the same. I would personally choose another name. Why not `convertToJsonPointerIfNeeded()` ?

jflabatBCSS (commented):
Some minor comments, but they are really optional.

belgif-rest-problem-validator/src/main/java/io/github/belgif/rest/problem/validation/AbstractRequestValidator.java
jpraet: true, but we'll actually remove the transformName call here https://github.com/belgif/rest-problem-java/pull/348#discussion_r3681478570 to rely on the auto-conversion instead.

If the user provides a correct name like "/ssins", we don't want to log an auto-conversion warning. To achieve that, we need to use the correct index format "/ssins/1" and not "/ssins[1]".

So as-is, the index format selection wasn't really needed, but in the final solution it will be needed.

jpraet (approved):

